PR #2, opened 2018-03-21
Testing

gdotdesign:
Language features for easy testing.


src/compilers/component.cr
```diff
@@ -90,14 +90,17 @@ class Compiler
     node.uses.each do |use|
       condition = use.condition ? compile(use.condition.not_nil!) : "true"
 
+      name =
+        underscorize(use.provider)
```
Sija: That is one weird name 😕 Why not just `underscore` (like in Rails' `ActiveSupport::Inflector.underscore`)?

gdotdesign: There is the `underscore` on String https://crystal-lang.org/api/0.23.1/String.html#underscore-instance-method but it's not converting dots `.` which I need. I didn't want to confuse the two, so I followed the format from rails `titelize` and `constantize`.

Sija: I see, Rails' `underscore` doesn't do it either, btw.

PR #5, opened 2018-05-13
Ameba v0.5.1 -> v0.6.0

veelenga:
This PR upgrades ameba. See https://github.com/veelenga/ameba/releases/tag/v0.6.0 for details.

The release is pretty new, so you can face some issues. I tried to reduce the false positiveness as much as possible, but if you find something, please report.

### DIFF
--- a/src/type_checkers/function.cr
+++ b/src/type_checkers/function.cr
@@ -28,8 +28,7 @@ module Mint
         return_type =
           check node.type
 
-        wheres =
-          check node.wheres
+        check node.wheres
 
         resolved =
           Comparer.compare(body_type, return_type)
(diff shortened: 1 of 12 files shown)


src/type_checkers/function.cr
```diff
@@ -28,9 +28,6 @@ module Mint
         return_type =
           check node.type
 
-        wheres =
-          check node.wheres
```
gdotdesign: The variable is not needed but the type check for the wheres is.

gdotdesign:
Thank you for this PR! :tada: 

There is one place which is a false positive,also the new features of Ambea seem great :+1:

veelenga:
Fixed

PR #7, opened 2018-05-25
Add a .gitignore file to the initial project scaffold

murielsilveira:
Ignore the `dist` folder because it's normally not add to the repositories.

Is the correct to ignore the `.mint` folder, I'm not sure?


gdotdesign:
> Is the correct to ignore the .mint folder, I'm not sure?

Yes it is correct :) Thank you :+1:

PR #9, opened 2018-05-27
add shard target

TeddyDD:
This change makes `shards build` works. Useful for curious people like me
who want to test Mint ;)


gdotdesign:
Thanks ;)

PR #13, opened 2018-05-28
Add debounce to live reload

Flaque:
# Overview
This fixes a bug where the live refresh can infinitely reload the page on Firefox. It might also happen on other browsers, though I'm not sure. 

# Changes Made
On the websocket's `onclose`, change `connect(true)` -> `setInterval(() => connect(true), 200);`


Flaque:
Done <3

gdotdesign:
Thanks :+1:

Issue #12, opened 2018-05-28
Add --auto-format flag to the start command


gdotdesign:
By default the developmet server should not format files when changed only when `--auto-format` flag is provided.

PR #14, opened 2018-05-29
feat: adding --auto-format flag for development server

### LINKED ISSUE #10: mint start modifies source.
Running `mint start` changes the sources files to use two space indenting every time I save my files. This really confuses the text editors redo-undo feature, undoing makes Sublime text use half two space indenting and half tab indenting for whatever reason.

I think letting tabs be a thing would help "developer happiness" as something that is being focused on because for me, and other people using non two space indenting that automatic change can be a tad annoying.

Dohxis:
### Feature
Adding `--auto-format` flag for a development server. By default its false, so files are not going to be formatted on save. Would love to get feedback as I don't have a lot of experience with Crystal.

Implements #12 
Fixes #10

### DIFF
--- a/src/commands/start.cr
+++ b/src/commands/start.cr
@@ -4,10 +4,14 @@ module Mint
       include Command
 
       define_help description: "Starts the development server."
+      define_flag auto_format : Bool,
+                  description: "Auto formats the source files when running development server.",
+                  default: false,
+                  required: false
 
       def run
         execute "Running the development server" do
-          Reactor.start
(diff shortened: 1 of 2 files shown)


gdotdesign:
The build failed because of a linting error, [ambea](https://github.com/veelenga/ameba) is used to check code against crystal style guide.

Otherwise it is looking good! :+1:

Dohxis:
Fixed the issues.

PR #15, opened 2018-05-29
feat: checking if port is used and adding new flag to change the default port

Dohxis:
### Feature
Added `--port` and `-p` flags so it would be easier to change the default port. Also added checking if a port is already used by a different application and in that case server will try to choose an open port.

ports

### DIFF
--- a/src/commands/start.cr
+++ b/src/commands/start.cr
@@ -8,10 +8,16 @@ module Mint
                   description: "Auto formats the source files when running development server.",
                   default: false,
                   required: false
+      define_flag port : Int32,
+                  description: "Change the port to serve the application on. (Default: 3000)",
+                  long: port,
+                  short: p,
+                  default: 3000_i32,
+                  required: false
 
       def run
         execute "Running the development server" do
-          Reactor.start flags.auto_format
+          Reactor.start flags.auto_format, flags.port
         end
       end
     end
--- a/src/commands/start.cr
+++ b/src/commands/start.cr
@@ -8,16 +8,22 @@ module Mint
                   description: "Auto formats the source files when running development server.",
                   default: false,
                   required: false
+      define_flag host : String,
+                  description: "Change the host to serve the application on. (Default: 127.0.0.1)",
+                  long: host,
+                  short: h,
+                  default: ENV["HOST"]? || "127.0.0.1",
+                  required: false
       define_flag port : Int32,
                   description: "Change the port to serve the application on. (Default: 3000)",
                   long: port,
                   short: p,
-                  default: 3000_i32,
+                  default: (ENV["PORT"]? || "3000").to_i,
                   required: false
 
       def run
         execute "Running the development server" do
-          Reactor.start flags.auto_format, flags.port
+          Reactor.start flags.host, flags.port, flags.auto_format
         end
       end
     end
--- a/src/utils/server.cr
+++ b/src/utils/server.cr
@@ -12,17 +12,45 @@ module Mint
       end
     end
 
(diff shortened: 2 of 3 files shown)


src/utils/server.cr
```diff
@@ -12,17 +12,45 @@ module Mint
...
+      else
+        new_port = config.port + 1
+        until port_open?(config.host_binding, new_port)
+          new_port = new_port + 1
+        end
+        terminal.print "#{COG} Port #{port} is used by a different application!\n"
+        terminal.print "#{COG} Would you like to to use port #{new_port} instead? (Y/n)\n"
+
+        use_new_port = gets
+        if use_new_port === "y" || use_new_port === "Y"
```
veelenga: Maybe a bit more uniform:

`use_new_port.downcase == "y"`

Dohxis: Fixed, thanks!

src/utils/server.cr
```diff
@@ -12,17 +12,45 @@ module Mint
...
       config.logger = Logger.new
       config.setup
 
-      server =
-        HTTP::Server.new(config.host_binding, config.port, config.handlers)
+      if port_open?(config.host_binding, port)
+        server =
+          HTTP::Server.new(config.host_binding, port, config.handlers)
+        terminal.print "#{COG} Development server started on http://#{config.host_binding}:#{port}/\n"
+      else
```
manveru: I think it should fail in this case. Having blocking terminal IO on a webserver isn't good practice since it looks like the server is running while in reality it's waiting for user input.
So we could at least check if the terminal is a tty first and fail otherwise. Something like `STDIN.tty?` should do the trick.

src/commands/start.cr
```diff
@@ -8,10 +8,16 @@ module Mint
                   description: "Auto formats the source files when running development server.",
                   default: false,
                   required: false
+      define_flag port : Int32,
+                  description: "Change the port to serve the application on. (Default: 3000)",
+                  long: port,
+                  short: p,
+                  default: 3000_i32,
+                  required: false
```
manveru: Also would be awesome if it respected the `PORT` and `HOST` environment variables.

src/utils/server.cr
gdotdesign: Yes @manveru is right, someone might use `mint start` as production server for whatever reason and suppressing the output like `mint start > /dev/null` would cause it to behave like it's running while in reality it's waiting for user input. 

This should be the branch of `elseif STDIN.tty?` and then the else branch would just exit with a nice message and a code of `1`.

Dohxis: Now I understand what are you talking about, going to add this today.

src/utils/server.cr
```diff
@@ -12,17 +12,48 @@ module Mint
...
+        terminal.print "#{COG} Port #{port} is used by a different application!\n"
+        terminal.print "#{COG} Would you like to to use port #{new_port} instead? (Y/n)\n"
+
+        use_new_port = gets
+        if !use_new_port.nil? && use_new_port.downcase == "y"
+          run(host, new_port)
+        else
+          terminal.print "#{COG} Exiting...\n"
+        end
+      else
```
gdotdesign: How failures are handled in mint is by raising an instance of error https://github.com/mint-lang/mint/tree/master/src/errors which the command handles https://github.com/mint-lang/mint/blob/master/src/commands/command.cr#L43 by printing it's message and exiting with code 1.

Would be nice to have it this way, but It can stay as is, it just needs to `exit 1` instead of returning `1`.

PR #37, opened 2018-06-17
Hindley Milner type checking algorithm

gdotdesign:
Replace naive type checking implementation with Hindley Milner one: https://en.wikipedia.org/wiki/Hindley–Milner_type_system

Based on: https://bitbucket.org/puffnfresh/js-type-inference/src/default/typeinference.js

This results in better type checking on complex type signatures.


src/utils/decoder.cr
```diff
@@ -54,6 +54,10 @@ module Mint
       id(node.name)
     end
 
+    def generate(node : TypeChecker::Variable)
+      raise "WTF"
```
Sija: debug leftover? ;)

gdotdesign: kind of ;) it's not finished yet :smile:

Issue #85, opened 2018-09-01
Add Contributing Guideline


wisn:
We need a contributing guideline with development instructions to avoid error when installing Crystal and so on. There a bit problem when I'm trying to install like SSL library error so it is better to give an instruction to a newcomer who want to contributing to Mint. I'm using Ubuntu 18.04 and I will send a PR for this.

PR #86, opened 2018-09-01
Add contributing guideline and Makefile development command

### LINKED ISSUE #84: Add Makefile command for development purpose
The current Makefile commands are `build`, `test`, and `documentation`. The `build` command compiles Mint and move it to `~/.bin/mint` which is the same name as the "stable" command name that I currently used. I think there should be another command for development purpose like compiles Mint and move it to `~/.bin/mint-dev`. The point is use different name so we could develop Mint compiler and Mint project with ease.

wisn:
Closes #84 and #85

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,49 @@
+# Contributing Guideline
+
+Read the general [Contributing guide][1], and then:
+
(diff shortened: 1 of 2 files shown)


gdotdesign:
Thanks :+1:

PR #88, opened 2018-09-02
Initialize project in current directory

### LINKED ISSUE #83: Prettify `mint init` error message
There is an exception raised when we are using `mint init` command without any path argument supplied.

screenshot from 2018-09-01 22-52-13

Also, here is when the `File exists`.

screenshot from 2018-09-01 23-04-55

DanielSokil:
Fixes #83

### DIFF
--- a/src/commands/init.cr
+++ b/src/commands/init.cr
@@ -10,13 +10,14 @@ module Mint
         required: false
 
       def run
-        if arguments.name
+        name = arguments.name || ""
+        if name.size > 0
           execute "Initializing a new project" do
-            Scaffold.run(arguments.name || ".")
+            Scaffold.run(name)
           end
         else
-          execute "Initializing a new project in current directory" do
-            Scaffold.run(".")
+          execute "Please provide project name" do
+            puts "Example: mint init my-project-name"
           end
         end
       end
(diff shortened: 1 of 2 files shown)


src/commands/init.cr
```diff
@@ -7,11 +7,17 @@ module Mint
 
       define_argument name,
         description: "The name of the new project",
-        required: true
+        required: false
```
Sija: I'd suggest leaving this as is since otherwise it might be too easy to initialize a project by accident in the wrong directory. Passing `.` is way more expressive.

DanielSokil: I can't seem to figure out how to handle if no `name` is passed.
https://github.com/jwaldrip/admiral.cr/

gdotdesign: We should change it to `false` and handle it manually by raising an error if the name is not present.

src/commands/init.cr
```diff
@@ -7,11 +7,17 @@ module Mint
...
 
       def run
-        execute "Initializing a new project" do
-          Scaffold.run(arguments.name)
+        if arguments.name
+          execute "Initializing a new project" do
+            Scaffold.run(arguments.name || ".")
+          end
+        else
+          execute "Initializing a new project in current directory" do
```
gdotdesign: Here we should raise an error with some nice text.

gdotdesign:
Thanks you :+1:

PR #89, opened 2018-09-05
Improve parser for block-comment

### LINKED ISSUE #81: Consider line comment
A line comment that is different from a block comment would be nice

1. just typing something like // at the start of the line is much more convenient than putting something at the front and the end 

2. Many editors have the concept of line and block comment - so my intellij plugin for example works a lot better with both.

The existing block comment is great `/* */`  

maybe a line comment could be `//`

wisn:
Why improve block-comment parser? The block-comment will breaks if we do these below.

```
/*!
 * Block comment
 */
```

```
/*
   Block comment until EOF
<<<EOF
```

Please review @gdotdesign


src/parsers/component.cr
```diff
@@ -7,7 +7,7 @@ module Mint
 
     def component : Ast::Component | Nil
       start do |start_position|
-        comment = self.comment
+        comment = self.comment || self.inline_comment
```
gdotdesign: There are a lot of places where the `comment` function is called. Adding `|| self.inline_comment` to each would not be a good solution.

Probably we should:
- rename the `comment` function to `block_comment`
- create a new `comment` function which parses a `block_comment` or an `inline_comment`
- rename the `Ast::Comment` node to `Ast::BlockComment` and create an alias: `alias Comment = BlockComment | InlineComment`

Then all current related code would work unchanged.

wisn:
Thanks for the review, @gdotdesign! Okay. I will update this PR. For the #81, it will be on separated PR.

gdotdesign:
Thank you :+1:

PR #93, opened 2018-09-13
[WIP] Add inline comment parser

wisn:
There is still some errors. From what I see, the problem is on the formatter.


wisn:
@gdotdesign I close this PR for now. Will do another contribution later.

PR #96, opened 2018-09-18
provide option to disable service workers on build

kingsleyh:
This provides an option on build to disable the service workers:

    mint build -s | --noServiceWorker

It does not disable service workers for the documentation server.

### DIFF
--- a/src/builder.cr
+++ b/src/builder.cr
@@ -1,6 +1,6 @@
 module Mint
   class Builder
-    def initialize(relative)
+    def initialize(relative, noServiceWorker)
       json = MintJson.parse_current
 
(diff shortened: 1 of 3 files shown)


Sija:
I'd go with positive flag `serviceWorkers` with default value `true`.

kingsleyh:
Hey @Sija I thought about this - but thought it might be strange to pass the flag `--serviceWorker` to turn it off - at least `--noServiceWorker` is very clear that it's turning it off

gdotdesign:
Thanks! :+1: 

I'll probably change the name of the flag to be in line with the rest `--skip-service-worker` also probably remove the short name because it's not relaying it's purpose.

PR #114, opened 2019-04-02
Optimize generated JavaScript code

gdotdesign:
This PR contains the code to minify the output files by mangling names and removing whitespace.

Also it contains some additional features:
- allow records to be created without type definition by creating a temporary type definition for them
- unify function call implementation - calls can be made on most expressions
- partial application of functions
- default properties are now only resolve once per component
- functions are bound to components and modules only once (in the constructor)
- unknown CSS properties are now result in an error


src/type_checkers/css_definition.cr
```diff
@@ -12,6 +18,13 @@ module Mint
             STRING
           end
 
+        if !node.name.starts_with?('-')
```
Sija:
```suggestion
        unless node.name.starts_with?('-')
```

gdotdesign: Thanks :+1:

PR #123, opened 2019-05-28
Update Http.mint

Acciaiodigitale:
I was reading the source and find this mismatch type in the comment......


gdotdesign:
Thank you very much! :+1:

PR #124, opened 2019-06-03
fix crash with remainder (%) operator

ezrast:
Currently using the % operator triggers a TypeError in `type_checkers/operation.cr`


gdotdesign:
Thank you!

Issue #128, opened 2019-06-11
Compose styles


gaku-sei:
First of all, let me tell that you I really, really, like this project, and hope it will reach v1 and be supported by a bigger community soon!

I have a question regarding the styles. Apparently, there is no way to use more than one style for an HTML node (`<div::foo::bar` is not allowed), and there is also no way to "inject" several css rules in a row (`style my-style { width: 20px; {moreStyle} }` throws a parse error), which leaves us with this "computed values" pattern:

```
style my-style {
  color: {color}
}

get color : String { ... }
```

Which is, in my humble opinion, pretty nice. Except when you need more than one or two computed values! The code gets longer and longer, and might become quite unsafe as well.

I tried to return a record, but it's still quite cumbersome, since anonymous records are apparently not supported in Mint, and therefore requires you to declare the record type additionally to the returned value.

So my question is; is there a better way to handle this use case? If not, is anything planned in the future?

The [roadmap](https://www.mint-lang.com/roadmap) says there will be some type checking for the css, meaning there will probably be some `Css` or `Style` type one day, making this: `get computedStyle : Css { ... }` possible.

gdotdesign:
Thank you for opening this issue!

A little inside on how things work in Mint:
[CSS declarations](https://developer.mozilla.org/en-US/docs/Web/CSS/Syntax#CSS_declarations) in Mint can be in three forms:
* **static** `background: red;` - these are compiled as is into a selector
* **semi-dynamic** `background: {some expression};` - these are compiled to use [CSS Variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) `background: var(--a);` 
* **dynamic** - currently, in the daily build you can set the `style` attribute of an element to either a `String` or a `Map(String, String)` like this: 
   ```
  <div::style style={Map.set("background", "red", Map.empty())}></div>
  ```
  and they are applied to the `style` attribute of the element.

The combining styles issue came up before as well: https://github.com/mint-lang/mint/issues/99 Personally I had only a few times that I felt that the language needs some way of combining styles however reading the description I had an idea on how it can be solved property.

What do you think about something like this:

```
style base {
  background: red;
  color: white;

  if (loading) {
    background: white;
    color: blue;
  }
}
```

Basically `if` statements would be allowed in style blocks (with or without an `else` branch) and can only contain CSS declarations. This would reduce the boilerplate as well as make the code more readable, and I think it can be implemented.

[...]

PR #140, opened 2019-10-23
[WIP] Refactor Styling

gdotdesign:
This PR contains the improvements and refactoring for the `style` tag. The related issue is #128

- [x] Remove ampersand from sub selectors
- [x] Change CSS interpolation to `#{...}` from `{...}`
- [x] Make selectors and media queries nestable
- [x] Make `if` available in `style`
- [x] Make `case` available in `style`
- [x] Allow parameters for `style`
- [x] Allow multiple styles for an element
- [x] Use fallback variables for CSS variables

## Breaking changes

- **The ampersand `&` is no longer required for sub selectors.** It is still required for pseudos. This was a mistake in my part in the first place, it just made it harder to understand what is going on. It was meant to be compatible with the [CSS3 Nesting Module](https://drafts.csswg.org/css-nesting/) which is still in draft state (after many years).

  This way it's more understandable and more compatible with preprocessors like SASS.

- **CSS interpolation is now `#{...}` instead of `{...}`.** This change was needed for two reasons:
  1. parsing CSS selectors and CSS definition using old syntax were in conflict (ambiguous) in certain cases (`a:hover {...}` vs `display: {...};`)), the ampersand `&` for the selectors was a solution for this but since it was removed for more clarity we need to prefix it.
  2. the old syntax was incompatible with JavaScript interpolation and the soon to be implemented string interpolation. All three will use the same syntax.

## New features

- **Selectors and media queries can be nested infinitely (even in each other)**. In the past it was not possible to nest media queries and selectors or even selectors in selectors, this is now possible:

  ```
  style base {
    max-width: 300px;

    div {
      color: red;

      span {
        color: blue;

        @media (max-width: 300px) {
          color: yellow;

          a {
            font-weight: bold;

            @media (max-height: 300px) {
              font-style italic;
            }
          }
        }
      }
    }
  }
  ```

  The example above will compile to:

  ```css
  .a {
    max-width: 300px;
  }

  .a div {
    color: red;
  }

  .a div span {
    color: blue;
  }

  @media (max-width: 300px) {
    .a div span {
      color: yellow;
    }

    .a div span a {
      font-weight: bold;
    }
  }

  @media (max-width: 300px) and (max-height: 300px) {
    .a div span a {
      font-weight: italic;
    }
  }
  ```

  Sub selectors will use the descendant operator and media queries will be combied using the `and` keyword.

- **`if` and `case` can be used inside style tags (in selectors and media queries as well)**. You can use them to apply styles conditionally:

  ```
  style base {
    background: red;
    color: white;

    if (loading) {
      background: white;
      color: blue;
    }

    case (status) {
      Status::Loading => color: blue;
      Status::Loaded => color: black;
      Status::Error => color: red;
    }
  }
  ```

  The `else` branch can be omitted in this case.

- **multiple styles can be added to the same element**, they are applied in order:
  
  ```
  component Main {
    style one {
      color: blue;
    }

    style two {
      backround: red;
    }

    style three {
      background: black;
    }

    fun render : Html {
      <div>
        <div::one::two>
        </div>

        <div::one::three>
        </div>
      </div>
    }
  }
  ```

- **arguments can now be specified for `style` blocks**:

  ```
  component Main {
    style item(color : String) {
      background: red;
      color: #{color};
    }

    fun render : Html {
      <div>
        <div::item("red")>
          "I am red!"
        </div>

        <div::item("blue")>
          "I am blue!"
        </div>
      </div>
    }
  }
  ```


s0kil:
The `if` statements in `style` will be evaluated at runtime or compile time?

gdotdesign:
The are evaluated at runtime into a [CSS Variable](https://www.w3schools.com/css/css3_variables.asp) this is an example of a compiled if https://github.com/mint-lang/mint/blob/refactor-styling/spec/compilers/css_with_if

PR #143, opened 2019-11-24
Caching component virtual doms 

gdotdesign:
This PR implements caching for components that are considered static.

Rendered virtual DOM nodes of components with static properties can be cached to increase performance and reduce compiled file size.

For example if we have an icon using SVG:
```
component Icons.Projects {
  property size : Number = 26

  fun render : Html {
    <Icon size={size}>
      <path
        d={
          "M96,176h80V96H96V176z M216,416h80v-80h-80V416z M96,416h8" \
          "0v-80H96V416z M96,296h80v-80H96V296z M216,296h80v-80h-80" \
          "V296z M336,96v80h80V96H336z M216,176h80V96h-80V176z M336" \
          ",296h80v-80h-80V296z M336,416h80v-80h-80V416z"
        }/>
    </Icon>
  }
}
```
And we use it in many places like this: `<Icons.Projects/>` it would re-render the virtual DOM every time but it's basically static content.

With this PR components uses that match the following criteria are cached:
* have no children
* does not have a reference attached to it
* all of it's attributes are considered static (number, bool, string literals or array containing those)

If you take the example above all of these uses will be cached:
*  `<Icons.Projects/>`
*  `<Icons.Projects size={30}/>`
*  `<Icons.Projects size={45}/>`

### DIFF
--- a/src/ast/array_literal.cr
+++ b/src/ast/array_literal.cr
@@ -8,6 +8,17 @@ module Mint
                      @from : Int32,
                      @to : Int32)
       end
+
+      def static?
+        items.all?(static?)
+      end
+
+      def static_value
+        values =
+          items.map(&.static_value).join(",")
+
+        "[#{values}]"
+      end
     end
   end
 end
--- a/src/ast/array_literal.cr
+++ b/src/ast/array_literal.cr
@@ -10,12 +10,12 @@ module Mint
       end
 
       def static?
-        items.all?(static?)
+        items.all?(&.static?)
       end
 
       def static_value
         values =
-          items.map(&.static_value).join(",")
+          items.map(&.static_value).join(',')
 
         "[#{values}]"
       end
--- a/src/ast/html_component.cr
+++ b/src/ast/html_component.cr
@@ -19,8 +19,8 @@ module Mint
 
       def static_hash
         component +
-          attributes.map(&.static_value).join("") +
-          children.map(&.static_value).join("")
+          attributes.map(&.static_value).join +
+          children.map(&.static_value).join
       end
     end
   end
--- a/src/ast/html_component.cr
+++ b/src/ast/html_component.cr
@@ -12,6 +12,14 @@ module Mint
                      @from : Int32,
                      @to : Int32)
       end
+
+      def static?
+        children.empty? && ref.nil? && attributes.all?(&.static?)
+      end
+
+      def static_hash
+        component + attributes.map(&.static_value).join("")
+      end
     end
   end
 end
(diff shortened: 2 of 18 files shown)


src/ast/array_literal.cr
```diff
@@ -8,6 +8,17 @@ module Mint
                      @from : Int32,
                      @to : Int32)
       end
+
+      def static?
+        items.all?(static?)
```
Sija:
```suggestion
        items.all?(&.static?)
```

src/ast/array_literal.cr
```diff
@@ -8,6 +8,17 @@ module Mint
...
+      end
+
+      def static_value
+        values =
+          items.map(&.static_value).join(",")
```
Sija:
```suggestion
          items.map(&.static_value).join(',')
```

src/ast/html_component.cr
```diff
@@ -12,6 +12,16 @@ module Mint
...
                      @to : Int32)
       end
+
+      def static?
+        children.all?(&.static?) && ref.nil? && attributes.all?(&.static?)
+      end
+
+      def static_hash
+        component +
+          attributes.map(&.static_value).join("") +
```
Sija:
```suggestion
          attributes.map(&.static_value).join +
```

src/ast/html_component.cr
```diff
@@ -12,6 +12,16 @@ module Mint
...
+          children.map(&.static_value).join("")
```
Sija:
```suggestion
          children.map(&.static_value).join
```

src/ast/html_element.cr
```diff
@@ -13,6 +13,19 @@ module Mint
...
+      def static?
+        children.all?(&.static?) &&
+          ref.nil? &&
+          attributes.all?(&.static?) &&
+          styles.empty?
+      end
+
+      def static_hash
+        component +
+          attributes.map(&.static_value).join("") +
```
Sija:
```suggestion
          attributes.map(&.static_value).join +
```

src/ast/html_element.cr
```diff
@@ -13,6 +13,19 @@ module Mint
...
+          children.map(&.static_value).join("")
```
Sija:
```suggestion
          children.map(&.static_value).join
```

src/ast/number_literal.cr
```diff
@@ -9,6 +9,18 @@ module Mint
...
+      def static?
+        true
+      end
+
+      def static_value
+        if float
+          value.to_s
+        else
+          value.to_i64.to_s
+        end
```
Sija:
```cr
(float ? value : value.to_i64).to_s
```

gdotdesign: I like `if` more so I went with this:
```
if float
  value
else
  value.to_i64
end.to_s
```

PR #154, opened 2020-01-08
Update Dependencies & Added Versioning

s0kil:
Does Crystal handle versioning similar to NPM, where Minor and Patch versions are always the latest available?


.gitignore
```diff
@@ -5,3 +5,4 @@ dist
 mint
 docs
 bin
+TODO.md
```
Sija: Why?

shard.yml
```diff
@@ -11,13 +11,13 @@ dependencies:
     version: 0.2.1
   baked_file_system:
     github: schovi/baked_file_system
-    version: 0.9.7
+    version: 0.9.8
```
Sija: It might be worth considering usage of `~>` operator instead of fixed versions.

.gitignore
s0kil: Done

.gitignore
```diff
@@ -4,4 +4,4 @@ mint-stuff
 dist
 mint
 docs
-bin
+bin
```
Sija: That change should be reverted.

shard.yml
gdotdesign: yeah probably we should do that :slightly_smiling_face:

gdotdesign:
Thanks :+1: and thanks for the review (as always) @Sija :wink:

Issue #151, opened 2020-01-06
External JavaScripts and CSS


gdotdesign:
To make it more easier to integrate with other JavaScript sources we should add the ability to add external CSS to the compiled code.

Alongside with this we should change it how it configured in the `mint.json` file, it should be like this:

```json
{
  "external": {
     "javascripts": ["path/to/javascript"],
     "css": ["path/to/css"]
   }
}
```

The provided CSS files should be checked the same way the JavaScripts are checked now and they should be added to the compiled CSS.

s0kil:
@gdotdesign Could we stick with 'javascript' over 'javascripts'?

gdotdesign:
I'm more comfortable with it's plural form since the user can add more than one and that conveys it more.

`css` is already in plural form `Cascading Style Sheets`

s0kil:
@gdotdesign I have got the CSS and JavaScripts (key, values) parsed, Where would the CSS be injected in the AST output?

gdotdesign:
I see two options:
- [here](https://github.com/mint-lang/mint/blob/master/src/compilers/top_level.cr#L108) is where the CSS gets compiled so I think it can be prepended to that
- during development we can serve them separately and during build time we can copy them over a reference them sepearately

The first option is easier to implement, but the second would be preferable (multiple smaller files instead of one big file)

s0kil:
> reference them separately

`<link rel="stylesheet" href="dist/styles-lib.css">`?

s0kil:
I was also thinking if an asset is a link to `http(s)://example.com/file.{js, css}` then fetch and cache it locally to project level `.mint` dir.

This would allow managing versions of external libs:
- `https://cdn.jsdelivr.net/npm/jquery@3.4.1/dist/jquery.min.js`
- `https://raw.githubusercontent.com/instantpage/instant.page/v3.0.0/instantpage.js`
- etc...

s0kil:
But that does not seem necessary, as a user could have Mint + NPM in the same project.

gdotdesign:
> > reference them separately
> 
> `<link rel="stylesheet" href="dist/styles-lib.css">`?

yes that was my thinking (maybe put it in a sub folder `dist/css/styles-lib.css`)

> I was also thinking if an asset is a link to `http(s)://example.com/file.{js, css}` then fetch and cache it locally to project level `.mint` dir.

[...]

PR #155, opened 2020-01-09
External JavaScripts and CSS

s0kil:
Implemented #151 
Please provide critical review.

Possible re-factors
`SourceFiles.external_files` should take advantage of type checking instead of `SourceFiles.external_files("javascripts")`


src/mint_json.cr
```diff
@@ -203,11 +206,28 @@ module Mint
...
+    # Parsing external assets (JavaScripts, CSS)
     # --------------------------------------------------------------------------
-    json_error MintJsonExternalJavascriptNotExists
+
+    json_error MintJsonExternalInvalid
+
+    def parse_external_assets
+      @parser.read_object do |key|
+        case key
+        when "javascripts"
```
Sija: I'd consistently use either form for both entries:
- `javascripts` + `stylesheets`
- `js` + `css`

s0kil: @gdotdesign Should we stick to `stylesheets` instead of `css`?

s0kil: `stylesheets` is nice, maybe in the future we could preprocess `sass` if required, the name is not limited to `css`.

`js` is not descriptive.

gdotdesign: yeah we should use `stylesheets` don't know why I didn't think of that :man_facepalming:

s0kil: More coffee is always the answer :smile:

src/utils/index_html.cr
```diff
@@ -56,6 +56,22 @@ module Mint
                   href: path_for("icon-#{size}x#{size}.png"))
               end
             end
+
+            unless json.external_files["css"].empty?
+              json.external_files["css"].each do |stylesheet|
+                # Assuming File Is In `public` Directory
```
gdotdesign: We don't need to assume because when parsing the `mint.json` we make sure that the file exists relative to the root directory (where the `mint.json` is).

There is two ways we can handle this:
- either we force the assets to be in the public directory which is reasonable, if so we need to check that when we parse them
- we use directory name which won't conflict with user paths, I am thinking about `.mint/css` and use that in both development and production, but then we need to implement the serving in the `Reactor`

I think probably forcing them to put assets like these in the public directory would be preferable, this would apply to the `javascripts` as well)

s0kil: Also, should we provide the option to link the javascripts as `script.src` vs being baked in with the application? In order to take advantage of browser caching, parallel execution, etc...

s0kil: We could also add a asset loader, which will load external `javascripts`.

s0kil: What I'm trying to convey is that Code Splitting is useful if we care about application start-up time, faster repeat load times, etc,

gdotdesign: I think service them as `<script src=""></script>` would be better, it can be an done in an other issue if like.

s0kil:
PR looks good to go.

PR #157, opened 2020-01-10
Handle Format Command Exception If Pattern Not Valid

s0kil:
```bash
$ mint format tests/
Mint - Formatting files
Unhandled exception: Error reading file: Is a directory (Errno)
Failed to raise an exception: END_OF_STACK
[0x564504d061a6] ???
[0x56450497e42b] __crystal_raise +43
[0x56450497ecd5] ???
[0x56450498b1ad] ???
[0x56450498373c] ???
[0x564504983163] ???
[0x564504a0c1b5] ???
[0x56450499765a] ???
[0x564504980eb7] main +55
[0x7ff3b65461e3] __libc_start_main +243
[0x564504901cde] _start +46
[0x0] ???
```

### DIFF
--- a/src/commands/format.cr
+++ b/src/commands/format.cr
@@ -10,33 +10,31 @@ module Mint
         default: "source/**/*.mint"
 
       def run
-        begin
-          execute "Formatting files" do
-            files = Dir.glob(arguments.pattern.to_s)
+        execute "Formatting files" do
+          files = Dir.glob(arguments.pattern.to_s)
 
-            if files.empty?
-              terminal.puts "Nothing to format!"
-            else
-              results =
-                files.map do |file|
-                  artifact =
-                    Parser.parse(file)
+          if files.empty?
+            terminal.puts "Nothing to format!"
+          else
+            results =
+              files.map do |file|
+                artifact =
+                  Parser.parse(file)
 
-                  formatted =
-                    Formatter.new(artifact, MintJson.parse_current.formatter_config).format
+                formatted =
+                  Formatter.new(artifact, MintJson.parse_current.formatter_config).format
 
-                  if formatted != File.read(file)
-                    File.write(file, formatted)
-                    terminal.puts "Formatted: #{file}"
-                    true
-                  end
-                end.compact
+                if formatted != File.read(file)
+                  File.write(file, formatted)
+                  terminal.puts "Formatted: #{file}"
+                  true
+                end
+              end.compact
 
-              terminal.puts "All files are formatted!" if results.empty?
-            end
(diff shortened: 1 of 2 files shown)


src/commands/format.cr
```diff
@@ -10,29 +10,33 @@ module Mint
...
+                    true
+                  end
+                end.compact
 
-            terminal.puts "All files are formatted!" if results.empty?
+              terminal.puts "All files are formatted!" if results.empty?
+            end
           end
+        rescue
+          print "I was looking for a valid pattern such \"source/**/*.mint\" \ngot \"#{arguments.pattern.to_s}\" instead.\n"
```
s0kil: Should this be in a new `Error` class? Similar to `SyntaxError`

src/commands/format.cr
```diff
@@ -10,29 +10,33 @@ module Mint
         default: "source/**/*.mint"
 
       def run
-        execute "Formatting files" do
-          files = Dir.glob(arguments.pattern.to_s)
+        begin
```
Sija: `begin ... end` is redundant here, `rescue` branch is enough.

src/commands/format.cr
Sija: `to_s` is redundant.

Sija: Since you rescue all of the exceptions how do you know it's the matter of incorrect pattern?

s0kil: We could have a seperate message if the pattern dirs/files do not exist.

s0kil: But this exception makes sense, if there are no files yielded from `Dir.glob` then the pattern is not valid.

s0kil:
Getting used to Git

src/commands/format.cr
```diff
@@ -33,6 +33,9 @@ module Mint
 
             terminal.puts "All files are formatted!" if results.empty?
           end
+        rescue
+          print "I was looking for a pattern that contains \".mint\" files,\n"
```
Sija:
```suggestion
          puts %(I was looking for a pattern that contains ".mint" files,)
```

src/commands/format.cr
```diff
@@ -33,6 +33,9 @@ module Mint
...
+          print "such as \"source/**/*.mint\". Got \"#{arguments.pattern}\" instead.\n"
```
Sija:
```suggestion
          puts %(such as "source/**/*.mint". Got "#{arguments.pattern}" instead.)
```

s0kil:
Looks Good To Go.

gdotdesign:
Thanks :+1:

PR #159, opened 2020-01-20
Format 'source' And 'test' Files

### LINKED ISSUE #158: Format skips test files
I love that this project formatting built-in! I'm using Mint `0.7.1` on macOS (installed today via brew), and the only files in my project are what I get with `mint-lang init <name>`.

As a test I completely out-dended both `source/Main.mint` and `tests/Main.mint`, then ran `mint-lang format`:

```
➤ mint-lang format **/*.mint
Mint - Formatting files
Formatted: source/Main.mint
All done in 1.367ms!
```

I also tried `mint-lang format source/** tests/**` and `mint-lang format {source,tests}/**`, but each only formatted `./source`.

I was only able to format `./tests` with `mint-lang format tests/**`, but I'd really prefer to format all Mint files in one go.

s0kil:
Fixes #158

### DIFF
--- a/src/commands/format.cr
+++ b/src/commands/format.cr
@@ -6,12 +6,15 @@ module Mint
       define_help description: "Formats source files"
 
       define_argument pattern,
-        description: "The pattern which determines which files to format",
-        default: "source/**/*.mint"
+        description: "The pattern which determines which files to format"
 
       def run
         execute "Formatting files" do
-          files = Dir.glob(arguments.pattern.to_s)
+          if arguments.pattern.to_s.empty?
+            files = Dir.glob(["tests/**/*.mint", "source/**/*.mint"])
+          else
+            files = Dir.glob(arguments.pattern.to_s)
+          end
 
           if files.empty?
             terminal.puts "Nothing to format!"
--- a/src/commands/format.cr
+++ b/src/commands/format.cr
@@ -10,8 +10,19 @@ module Mint
 
       def run
         execute "Formatting files" do
+          current =
+            MintJson.parse_current
+
+          format_directories =
+            current.source_directories | current.test_directories
+
+          format_directories_patterns =
+            format_directories.map do |dir|
+              File.join(dir, "**/*.mint")
+            end
+
           if arguments.pattern.to_s.empty?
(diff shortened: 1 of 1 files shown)


s0kil:
Should we respect the `source-directories` and `test-directories` configuration?

gdotdesign:
I think it's a good idea, certainly better than using hard coded paths.

s0kil:
It looks good to go.

src/commands/format.cr
```diff
@@ -6,12 +6,26 @@ module Mint
...
+          format_directories_patterns =
+            format_directories.map do |dir|
+              File.join(dir, "**/*.mint")
+            end
+
+          if arguments.pattern.to_s.empty?
+            files = Dir.glob(format_directories_patterns)
+          else
+            files = Dir.glob(arguments.pattern.to_s)
+          end
```
Sija: One pedantic detail - what about this:

```cr
file_pattern = arguments.pattern.to_s.presence || format_directories_patterns
files = Dir.glob(file_pattern)
```

s0kil: I like that. What is `presence` in this context?

Sija: See [`String#presence`](https://crystal-lang.org/api/0.32.1/String.html#presence:self?-instance-method) and [`Nil#presence`](https://crystal-lang.org/api/0.32.1/Nil.html#presence-instance-method)

s0kil: Thank You, What is the relationship with `?`?

Sija: I'm sorry, I don't get the question...

Sija: I'm sorry, I don't get the question...

s0kil: Does it make sense to do just `presence` or `present?` ?

Sija: `presence` is the name.

s0kil: Got it.

s0kil: Tried it, turned into a crazy error message, ha
https://pastebin.com/fkjgJ2r4

s0kil: I don't want to spend much time on something so insignificant.

gdotdesign:
Thanks :+1:

Issue #179, opened 2020-02-12
Feature Request: Recursion for local functions


Namek:
A function defined in the `where` block cannot call itself. Compiler properly detects the situation but errors with:
 > Recursion is not supported at this time by the language.

It's a common pattern to call a helper function that receives more arguments than a module function. I think we should have it. What are the obstacles?

gdotdesign:
There is no obstacle here, since normal function recursion is solved by using it's static type definition, we can do the same thing here, just need to code it.

If anyone wants to give this a go before I can get to it, here are some pointers:
- here (https://github.com/mint-lang/mint/blob/master/src/type_checker.cr#L223) is where the type-checker does the check for recursion, we need to check for `WhereStatement`
- here (https://github.com/mint-lang/mint/blob/master/src/type_checkers/function.cr#L6) is the `static_type_signature` of a function, we need to do the same for `WhereStatement`

Namek:
Can I request a release with this? Thanks.

gdotdesign:
Sure, I'll try to do a release in a few days.

PR #181, opened 2020-02-14
allow recursive calls in/to inline functions

Namek:
https://github.com/mint-lang/mint/issues/179

### DIFF
--- a/src/type_checker.cr
+++ b/src/type_checker.cr
@@ -222,6 +222,15 @@ module Mint
               return NEVER.as(Checkable)
             elsif node.is_a?(Ast::Function)
               static_type_signature(node)
+            elsif node.is_a?(Ast::WhereStatement)
+              expression =
+                node.expression
+
+              if expression.is_a?(Ast::InlineFunction)
+                static_type_signature(expression)
+              else
+                resolve expression
+              end
             else
               raise Recursion, {
                 "caller_node" => @stack.last,
--- a/spec/compilers/inline_function_recursive
+++ b/spec/compilers/inline_function_recursive
@@ -1,0 +1,44 @@
+module Test {
+  fun factorial(n : Number) : Number {
+    helper(n, 1)
+  } where {
+    helper = (n : Number, acc : Number) : Number {
+      if (n == 0) {
+        acc
+      } else {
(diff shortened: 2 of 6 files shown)


src/type_checker.cr
```diff
@@ -222,6 +222,8 @@ module Mint
               return NEVER.as(Checkable)
             elsif node.is_a?(Ast::Function)
               static_type_signature(node)
+            elsif node.is_a?(Ast::WhereStatement)
+              static_type_signature_for_inline_function(node)
```
gdotdesign: We can keep `static_type_signature` method since crystal has overloading https://crystal-lang.org/reference/syntax_and_semantics/overloading.html

src/type_checkers/where_statement.cr
```diff
@@ -1,5 +1,26 @@
 module Mint
   class TypeChecker
+    def static_type_signature_for_inline_function(node : Ast::WhereStatement) : Checkable
+      expression =
+        node.expression
+
+      if expression.is_a?(Ast::InlineFunction)
```
gdotdesign: The code for calculating  the static type signature of the `Ast::InlineFunction` should be in its type-checking file, because we might use it later.

src/type_checker.cr
Namek: I named it like that because it's not a type signature for WhereStatement but for InlineFunction IF it appears to be inside WhereStatement.

gdotdesign: Then maybe we should move checking the condition here? 

```
elsif node.is_a?(Ast::WhereStatement) && node.expression.is_a?(Ast::InlineFunction)
  static_type_signature(node)
````

I'm not sure if the compiler is smart enough to handle this.

src/type_checkers/where_statement.cr
Namek: I moved it to `inline_function.cr`

src/type_checker.cr
Namek: Nope, it is not.

Namek:
@gdotdesign please re-review :)

gdotdesign:
Thanks :+1: merged :rocket:

PR #183, opened 2020-02-15
Compile external assets into their own files.

gdotdesign:
This PR resolves issues with packages having external files:
- external javascripts are now compiled into a single file `external-javascripts.js` (in development and build)
- external stylesheets are now compiled into a single file `external-stylesheets.css` (in development and build)
- external files are no longer being restricted to the `public` directory

With the change of #155 the following issues cropped up:
- packages could not serve their external stylesheets in development because we relied on them being in the `public` directory
- having restrict external files into the public directory made it impossible to reference files outside of the project. To give this more context: I'm working on `mint-ui` and `mint-ui-showcase`, I needed to access a JavaScript file which is `mint-ui` from `mint-ui-showcase` as (`../mint-ui/assets/scroll_into_view.js`) and this was not possible

### DIFF
--- a/src/reactor.cr
+++ b/src/reactor.cr
@@ -104,13 +104,13 @@ module Mint
       get "/external-javascripts.js" do |env|
         env.response.content_type = "application/javascript"
 
-        SourceFiles.external_javascripts
+        @watcher.external_javascripts.to_s
       end
 
       get "/external-stylesheets.css" do |env|
         env.response.content_type = "text/css"
 
-        SourceFiles.external_stylesheets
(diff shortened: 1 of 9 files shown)


src/reactor.cr
```diff
@@ -101,6 +101,18 @@ module Mint
         script
       end
 
+      get "/external-javascripts.js" do |env|
+        env.response.content_type = "application/javascript"
+
+        SourceFiles.external_javascripts
```
Sija: I'd cache that, ditto below.

PR #189, opened 2020-02-16
Fix formatting

KevinSjoberg:
Running `crystal tool format` on the project revealed some files not being correctly formatted.

### DIFF
--- a/spec/parsers/string_literal_spec.cr
+++ b/spec/parsers/string_literal_spec.cr
@@ -11,7 +11,7 @@ describe "String Literal" do
   expect_error %("asd" \\), Mint::Parser::StringExpectedOtherString
 
   expect_ok %("OK")
-  expect_ok %("\"OK")
+  expect_ok %(""OK")
   expect_ok %("OK"\\"OK")
   expect_ok %("OK" \\ "OK")
 end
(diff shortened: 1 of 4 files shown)


src/utils/service_worker.cr
```diff
@@ -91,8 +91,8 @@ module Mint
         .glob("**/*")
         .reject { |file| File.directory?(file) }
         .reduce(OpenSSL::Digest.new("SHA256")) do |digest, file|
-          digest.update File.read(file)
-        end.to_s
+        digest.update File.read(file)
+      end.to_s
```
Sija: That seems wrong... might a formatter bug.

gdotdesign: @KevinSjoberg did you use Crystals formatter here?

KevinSjoberg: @gdotdesign yes, I ran `crystal tool format`.

KevinSjoberg: Just noticed I was a couple of versions behind. I was using `0.30.1`. Upgrading to `0.33.0` now.

KevinSjoberg: I've updated the pull request with the changes from running `crystal tool format` using the latest version of Crystal (0.33.0).

PR #190, opened 2020-02-16
Add Math.random

KevinSjoberg:
I was not sure if we were missing `Math.random` or if there is valid reason to why we're not supporting it yet. Feel free to close this pull request with an explanation if that's the case. 🙂

### DIFF
--- a/core/source/Math.mint
+++ b/core/source/Math.mint
@@ -117,4 +117,11 @@ module Math {
         to * 100
       }
   }
+
+  /*
+  Returns a pseudo-random number in the range 0 to less than 1.
+  */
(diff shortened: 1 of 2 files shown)


Namek:
Personally, I would be after a `Random` module that would support deterministic results based on seeds. Of course, we need also the `Math.random()` to somehow produce a randomized seed but maybe better in the other module rather than `Math`. I would love to hear your inputs!

gdotdesign:
I think `Math.random` have it's use cases and there is no reason why there cannot be a `Random` module that supports a seed.

Thanks for the PR :+1:

PR #194, opened 2020-02-17
Constants

### LINKED ISSUE #169: Is there a way to define global constants?
Sorry if I'm missing anything, but I can't seem to figure out a way to define global constants. In JavaScript, I can define a constant in a file like:  
```javascript
export const theme = {
  primaryColor: '#123456',
};
```
or
```javascript
export const CATEGORIES = ['cat1', 'cat2', 'cat3'];
```
How do I achieve this in `mint`? Thank you!

gdotdesign:
This PR will add constants to the language (and resolve #169).

Constants can be added to a  `store`, `module` or `component` using the following syntax:

```
module Math {
  constant PI = 3.141592653589793
}
```

- constants can be accessed like variables inside the scope of the entity which defined them
- constants can be accessed from the outside using the following syntax `Math:PI`

At this point I refrained from adding global constants, it can be added later if there is need.


MrCreeper1008:
I still feel like the single colon syntax is a bit confusing..

Namek:
I'd prefer `const`. It's typical for most languages and involves a bit less typing.

gdotdesign:
> how about `Module.myContant`

With the current implementation we cannot do that because it conflicts with type ids (since they start with an uppercase letter and can contain dots `.`).

> I'd prefer `const`. It's typical for most languages and involves a bit less typing.

I'm fine with that.

PR #195, opened 2020-02-17
Ignore .dwarf files

KevinSjoberg:
Whenever I run `make test-core` it generates a `mint.dwarf` files that I have to manually remove. I assume it's an artifact from building `mint` to run the core tests.

I guess we could also manually remove it as an extra step in the `Makefile`. Thoughts?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ dist
 mint
 docs
 bin
+*.dwarf
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ test:
 	crystal spec -p --error-trace && bin/ameba
 
 test-core:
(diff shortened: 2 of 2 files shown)


gdotdesign:
> I guess we could also manually remove it as an extra step in the Makefile. Thoughts?

Yeah it's fine to delete that (it's the side effect of compiling a binary https://github.com/crystal-lang/crystal/pull/4969), I don't have this since I work on Linux.

Also it gets uploaded to the releases which I have to manually remove.

KevinSjoberg:
@gdotdesign I've pushed a commit that adjusts the `Makefile` so that it will remove the `mint.dwarf` file if present.

PR #197, opened 2020-02-17
core: add Maybe.andThen()

Namek:
Useful for chaining value transformations without nesting multiple ifs. Railway Oriented Programming: https://youtu.be/vDe-4o8Uwl8?t=2662


gdotdesign:
Thanks :+1:

Issue #147, opened 2019-12-09
mint-0.7.0-osx show version 0.6.0


Acciaiodigitale:
Hi @gdotdesign.

When I'm updating mint for homebrew I noticed that 

`
mint-lang version
`

still show 0.6.0

s0kil:
@gdotdesign Forgot to update `shard.yml` (:

Acciaiodigitale:
I recompiled from source: now homebrew version show already the right version

gdotdesign:
Yes I forgot to update, I just released 0.7.1 for the new Crystal version and I didn't forgot to bump the version this time :sweat_smile:

PR #198, opened 2020-02-18
Enforce mint version

kevinsjoberg:
This will ensure new applications generated using `mint init` include the `mint-version` field. Whenever `mint install` is run the version specified in `mint-version` will be checked against the version of Mint you're running.

Upon mismatch the following error is presented:

Screenshot 2020-02-18 at 21 42 45

I removed the macro altogether for grabbing the version number from the `shard.yml` file since the test I've added will fail if those mismatch anyway.


s0kil:
@KevinSjoberg Instead of enforcing the Mint version, what about adding a warning?
If the version is enforced, it is always necessary to update `mint.json` even if there are no breaking changes.

It would be even better to show this warning when are error happens,
"Error Happened, possiblity beacuse of a version mismatch"

Sija:
Also, it needs to accept semver version selectors to make it flexible enough to use.

src/version.cr
```diff
@@ -0,0 +1,3 @@
+module Mint
+  VERSION = "0.8.0-devel"
```
Sija:
```suggestion
  VERSION = {{ `shards version #{__DIR__}`.chomp.stringify }}
```

Namek:
Should be totally optional and non-default in scaffold. It makes it inconvenient to develop the compiler this way.

src/version.cr
s0kil: Sometimes Mint gets a new version released without the updated version in the CLI, we do not want to make this problem worse. As @Sija said, one place to define the version is a good idea.

KevinSjoberg: @s0kil could you give me an use case for when that could happen? I'm fine with using the suggested compile-time macro by @Sija if that's the better choice. @Sija will that work cross-platform?

gdotdesign: @Sija I was thinking of using the [`read_file`](https://crystal-lang.org/api/0.33.0/Crystal/Macros.html#read_file(filename):StringLiteral-instance-method) macro, is `shards version` better?

gdotdesign:
I agree with @Sija that it ideally it should be a semver constraint (like in a dependency) or we take version given as the lowest supported version (`> 0.7.1`).

KevinSjoberg:
Thanks for the feedback. Really appreciate it. 🙂

The `mint-version` field was pretty much directly taken from Elm. The reason I wanted to introduce it was due to the fact that the Mint version of [Real World](https://github.com/mint-lang/mint-realworld) didn't run due to changes made in recent versions of Mint.

@Sija I do like the idea of a semver constraint but that will only work as long as we're doing proper versioning. Technically there shouldn't be any breaking changes between `0.8.0-devel` and `0.7.1` but it still broke the RealWorld app.

@Namek I do think it should be part of the scaffold but the mechanics of how we're enforcing `mint-version` should perhaps be adjusted. For instance, if the `mint-version` is a semver constraint it would be much less annoying but still very useful.

src/version.cr
Sija: @gdotdesign `shards version` exists for this exact reason, what's the point of using `read_file` instead?

@KevinSjoberg yes, it is (as long `shards` is).

Sija:
@KevinSjoberg SemVer works like that in general but IMO it's still way better than having fixed version that needs to updated every, damn, time (PITA).

src/version.cr
gdotdesign: I guess my only concern if it will work on windows builds.

KevinSjoberg:
@Sija yeah, I agree that having a SamVer constraint is flexible enough to not be painful but still useful enough that it's worth having. Perhaps a reasonable default in the scaffold would be `~0.8`, e.g. `>= 0.8, < 1`?

src/version.cr
Sija: It should, of course. If it ain't, that would mean there is a bigger problem than that (`shards`). Also, Windows support is still in its infancy, we can always fix things l8r if deemed necessary.

s0kil: @KevinSjoberg https://github.com/mint-lang/mint/issues/147

KevinSjoberg: @s0kil Ah, shit happens. 🙂

Sija:
@KevinSjoberg Are you still interested in finishing this?

If so...

> @Sija yeah, I agree that having a SamVer constraint is flexible enough to not be painful but still useful enough that it's worth having. Perhaps a reasonable default in the scaffold would be `~0.8`, e.g. `>= 0.8, < 1`?
>
> That would mean that whenever you run `mint init` the `mint-version` will be set to a constraint enforcing the version to be equal to or larger than current major, minor and up to but not including next major?

Make sense, let's do that 👍

KevinSjoberg:
@Sija sorry, I've been swarmed recently. I will probably not have chance to finish this in the near future.

gdotdesign:
@Sija I've finished this, will merge after review :sunglasses:

src/mint_json.cr
```diff
@@ -40,7 +41,13 @@ module Mint
           "node" => node(exception),
         }
       end
+
       parse_root
+
+      # TODO: Enable when releasing 0.13.0!
```
gdotdesign, line 47: Only enable this until `0.13.0` so it would not cause issues with testing things until then.

src/mint_json.cr
```diff
@@ -154,6 +163,64 @@ module Mint
...
+
+          Installer::SimpleConstraint.new(lower, upper)
+        else
+          raise MintJsonMintVersionInvalid, {
+            "node" => node(location),
+          }
+        end
+
+      resolved =
+        Installer::Semver.parse(VERSION.rchop("-devel")).not_nil!
```
gdotdesign, line 209: Semver parsing would fail with `-devel`.

src/mint_json.cr
```diff
@@ -40,7 +41,13 @@ module Mint
           "node" => node(exception),
         }
       end
+
       parse_root
+
+      # TODO: Enable when releasing 0.13.0!
+      # raise MintJsonMintVersionMissing, {
+      #   "contents" => @json,
+      # } unless mint_version_parsed
```
Sija, line 50: I'd allow for missing `mint-version` and treat it as `*`, i.e. any version. In this way it'll be enforced only when set, giving an "escape hatch" when you don't want to specify the `mint-version`.

```suggestion
```

gdotdesign: Fair enough :slightly_smiling_face:

kevinsjoberg:
Nicely done! 🎉

PR #202, opened 2020-02-19
Integrate with Preact-only runtime

KevinSjoberg:
This is a first attempt to make Mint compatible with the `preact` Mint runtime (meaning no `preact/compat`).

See https://github.com/mint-lang/mint-runtime/pull/8 for details.


core/source/Test/Html.mint
```diff
@@ -10,9 +10,8 @@ module Test.Html {
     (() => {
       let root = document.createElement('div')
       document.body.appendChild(root)
-      ReactDOM.render(#{node}, root)
+      _render(#{node}, root)
       return new TestContext(root, () => {
-        ReactDOM.unmountComponentAtNode(root)
```
KevinSjoberg: Not too sure about this. I mentioned something similar in https://github.com/mint-lang/mint-runtime/pull/8#issuecomment-588424638.

gdotdesign:
@KevinSjoberg is there something to do here? The unmount thing is the only one thing that still needs to be resolved?

KevinSjoberg:
@gdotdesign still waiting for feedback on the runtime pull request and so I’ve waited with any changes necessary here. I’ll make sure to update this as well whenever the runtime pull request is ready. 🙂

core/source/Test/Html.mint
gdotdesign: We need to keep this functionality somehow so tests are run in a clean state: components needs to unsubscribe from providers and stores (which runs in `componentWillUnmount`) does Preact has something which can do this?

gdotdesign: I think I found out how they do it: https://github.com/preactjs/preact/blob/master/compat/src/index.js#L67

gdotdesign:
I've kept the original interface so only the runtime changed :arrow_right:  https://github.com/mint-lang/mint/commit/dc31b6a412377b09d627bd17b3a66c3cb58cefa2

Issue #205, opened 2020-02-20
`mint test` in Docker: no --host option


Namek:
Since I use Windows I have to improvise with Mint.

`mint start` can be started within a Docker container but the server won't be available outside since 127.0.0.1 is container's own adress. With the `-h 0.0.0.0` option the served website becomes available.

However, we can't do that with `mint test`. There is no `host` option.

BTW this is how I tried to run it:
```Dockerfile
FROM zenika/alpine-chrome:latest

USER root
# RUN wget -O mint https://github.com/mint-lang/mint/releases/download/0.7.1/mint-0.7.1-linux
ADD mint /mint # add locally built version
RUN chmod +x /mint

EXPOSE 3001
WORKDIR /proj
ENTRYPOINT ["/mint"]
```

`docker run --tty -v c:/dev/Projects/my_project:/proj -p 3001:3001 --rm -ti mint test`

Namek:
Solved!

PR #207, opened 2020-02-20
Make hostname:port configurable for test server.

Namek:
Solves #205 

## What?

There are 4 new options for `mint test`:
- host
- port
- browser_host
- browser_port

## Running test using Docker:

`mint test --manual  --host 0.0.0.0`

This launches test server on host `0.0.0.0` which is available outside of the Docker container but also sets `--browser-host` to `127.0.0.1` by default which is a connection address for websocket outside the container. Same with `port`/`browser_port` pair.

Full instruction will be at https://github.com/mint-lang/mint-docker/issues/1

### DIFF
--- a/src/commands/test.cr
+++ b/src/commands/test.cr
@@ -9,7 +9,7 @@ module Mint
         short: "m"
 
       define_flag browser : String,
-        description: "Which browser to run the tests in",
+        description: "Which browser to run the tests in (chrome, firefox)",
         default: "chrome",
         short: "b"
 
@@ -17,6 +17,30 @@ module Mint
         description: "Which reporter to use (dot, documentation)",
         default: "dot",
         short: "r"
+
+      define_flag host : String,
+        description: "Host to serve the tests on. (Default: 127.0.0.1)",
+        default: ENV["HOST"]? || "127.0.0.1",
+        required: false,
+        short: "h"
+
+      define_flag port : Int32,
+        description: "Port to serve the tests on. (Default: 3001)",
+        default: (ENV["PORT"]? || "3001").to_i,
+        required: false,
+        short: "p"
+
+      define_flag browser_host : String,
+        description: "Target host, useful when hosted on another machine. (Default: 127.0.0.1)",
+        default: ENV["BROWSER_HOST"]? || "127.0.0.1",
+        required: false,
+        short: "x"
+
+      define_flag browser_port : Int32,
+        description: "Target port, useful when hosted on another machine. (Default: 3001)",
+        default: (ENV["BROWSER_PORT"]? || "3001").to_i,
+        required: false,
+        short: "c"
 
       define_argument test : String
 
--- a/src/test_runner.cr
+++ b/src/test_runner.cr
@@ -8,7 +8,8 @@ module Mint
       })
     end
 
-    PAGE = <<-HTML
+    def page_source : String
+      @page_source ||= <<-HTML
       <html>
         <head>
         </head>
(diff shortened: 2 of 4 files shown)


src/test_runner.cr
```diff
@@ -110,6 +111,7 @@ module Mint
         </body>
       </html>
     HTML
```
Sija:
```suggestion
      HTML
```

src/test_runner.cr
```diff
@@ -8,7 +8,8 @@ module Mint
       })
     end
 
-    PAGE = <<-HTML
+    def get_page
```
Sija:
```suggestion
    def page_source : String
```

src/test_runner.cr
```diff
@@ -282,7 +285,7 @@ module Mint
       get "/" do
         @failed = [] of Message
         @succeeded = 0
-        PAGE
+        get_page
```
Sija:
```suggestion
        page_source
```

src/test_runner.cr
```diff
@@ -8,7 +8,8 @@ module Mint
       })
     end
 
-    PAGE = <<-HTML
+    def page_source : String
+      <<-HTML
```
Sija:
```suggestion
      @page_source ||= <<-HTML
```

Namek:
Please don't merge it until my signal. I want to make it work on Windows first.

Namek:
@Sija @gdotdesign It works, I have also updated the PR description. Please re-review and merge if it's OK.

src/test_runner.cr
```diff
@@ -18,7 +19,7 @@ module Mint
           <script>
             class TestRunner {
               constructor (suites) {
-                this.socket = new WebSocket("ws://localhost:3001/")
+                this.socket = new WebSocket("ws://#{@flags.browser_host}:#{@flags.port}/")
```
Sija: Why not just set it to `document.location.origin`?

Namek: Because Docker might expose other port.

gdotdesign: I don't understand why we we use the `port` flag instead of the `browser_port` flag? 

My understanding is for example: in docker the server is we bound to `0.0.0.0:3000` which is exposed as `127.0.0.1:4000` then the `host` is `0.0.0.0`, `port` is `3000`, `browser_host` is `127.0.0.1` and `browser_port` is `4000`

Namek: Ohh, you're absolutely right! A typo. I'll fix it when I'll get back to PC

Namek: @gdotdesign done

Sija: @Namek is it?

src/commands/test.cr
```diff
@@ -18,6 +18,30 @@ module Mint
...
+        description: "Host to serve the tests on. (Default: 127.0.0.1)",
+        default: ENV["HOST"]? || "127.0.0.1",
+        required: false,
+        short: "h"
+
+      define_flag port : Int32,
+        description: "Port to serve the tests on. (Default: 3001)",
+        default: (ENV["PORT"]? || "3001").to_i,
+        required: false,
+        short: "bp"
```
Sija: Invalid, has to be a single char.

src/commands/test.cr
```diff
@@ -18,6 +18,30 @@ module Mint
...
+
+      define_flag browser_host : String,
+        description: "Target host, useful when hosted on another machine. (Default: 127.0.0.1)",
+        default: ENV["BROWSER_HOST"]? || "127.0.0.1",
+        required: false,
+        short: "bh"
```
Sija: ditto

src/commands/test.cr
Namek: any proposition?

Sija: I guess any letter will do, `x`? ;)

gdotdesign:
Thanks :+1:

Issue #193, opened 2020-02-16
Feature Request: Tuples


Namek:
It would be nice to have tuples such as pairs, triples and so on. I am not sure if that should be syntax thing or just a module (however then, with #191 done).

A dependency to this may be having **hole types in records** (mentioned in #130) or just a built-in module similar to `Array`, `Map` or `Set`.

## Why do we need a Tuple?

Sometimes we want to return a more complicated result in code of general usage, like packages.

For example, I have written myself a function like this in Elm:

```elm
{-| Find an element and return it along with the rest of the input items.
-}
findAndGetRest : (a -> Bool) -> List a -> ( Maybe a, List a )
findAndGetRest predicate list =
    let
        iterate checked unchecked =
            case unchecked of
                [] ->
                    ( Nothing, list )

                first :: rest ->
                    if predicate first then
                        ( Just first, List.append (List.reverse checked) rest )

                    else
                        iterate (first :: checked) rest
    in
    iterate [] list
```

which, I believe, is a good candidate for some publicly shared package.

I could have more functions returning tuples and for each one I would may have to create possibly a separate record with some unnecessary names.

## Current solution

Currently it can be done using a phantom enum/type.

```
enum FirstRest(a) {
  A(Maybe(a), Array(a))
}

module Util.Collection {
  /* Find an element and return it along with the rest of the input items. */
  fun findAndGetRest (predicate : Function(a, Bool), items : Array(a)) : FirstRest(a) {
    iterate([], items)
  } where {
    iterate =
      (checked : Array(a), unchecked : Array(a)) : FirstRest {
        try {
          case (unchecked[0]) {
            Maybe::Nothing => FirstRest::A(Maybe::Nothing, [])

            Maybe::Just first =>
              try {
                rest =
                  Array.slice(1, Array.size(unchecked))

                if (predicate(first)) {
                  FirstRest::A(Maybe::Just(first), Array.push(checked, rest))
                } else {
                  iterate(Array.push(first, checked), rest)
                }
              }
          }
        }
      }
  }
}
```

## How?

Tuple syntax is cool but it's not that simple to implement. Very easy solution would be to have a `Tuple` module, supporting record constructors. However, it is so easy to write one byself in a project, that it's no need to put it into API until we'll be sure what we need. In the other hand, writing `Tuple.pair(a, b)` each time instead of `{a, b}` or `(a, b)` (or w/e syntax) is not handsome.

gdotdesign:
[...]

PR #209, opened 2020-02-24
Tuples

gdotdesign:
This PR implements [**tuples**](https://en.wikipedia.org/wiki/Tuple) for the language #193. Tuples are fixed fixed sets of values which can be different types. Tuples are represented internally as arrays.

Tuples can be destructured in three places:
- `try`, `sequence` and `parallel` expressions
- `where` statements
- `case` statements

The syntax for a tuple looks like this: 
```
{"A", 0, true}
```
They resolve into the `Tuple` type where the parameters are types of it's members, for the example above:
```
Tuple(String, Number, Bool)
```

-----------------------

`Ast::Tuple`:
- [x] Parser
- [x] Type Checker 
- [x] Formatter
- [x] Compiler

`try`, `sequence` and `parallel` destructuring:
- [x] Modify `Ast::Statement` to handle destructuring
- [x] Modify `Ast::Try`, `Ast::Sequence`, `Ast::Parallel` to resolve destructuring
- [x] Raise error when trying to destructure more variables than available
- [x] Raise error when trying to destructure a none `Tuple` 

`where` destructuring:
- [x] Modify `Ast::WhereStatement` to allow for destructuring
- [x] Raise error when trying to destructure more variables than available
- [x] Raise error when trying to destructure a none `Tuple` 

`case` destructuring:
- [x] Add `Ast::TupleDestructuring`
  - [x] Parser
  - [x] Type Checker 
  - [x] Formatter
- [x] Modify `Ast::CaseBranch` to resolve tuple destructuring
- [x] Raise error when trying to destructure a none `Tuple`


gdotdesign:
This is pretty much done :tada: @Namek can you test it if it fits your use-cases?

gdotdesign:
@Sija would you mind review it? :slightly_smiling_face:

Namek:
First test: it works. But I will have more time on weekend to work with this some more. 

This might look weird:
```
if (condition) {
  { x + 1, y }
} else {
  { x, y + 1 }
}
```

with one-liner formatting:

```
if (condition) { { x + 1, y } } else { { x, y + 1 } }
```

gdotdesign:
I've udpated the PR with some refactoring with the additional change that destructuring needs to be wrapped in `{}` like this:

```
{a, b, c} => {"A, "B", "C"}
```

This is to match the array destructuring semantics #210

Sija:
Same https://github.com/mint-lang/mint/pull/210#issuecomment-592022688 applies here.

Namek:
The record update syntax may behave wrong. I wrote code as such:

```
{genState |
  fieldsIndices = Array.push(idx, genState.fieldsIndices),
  fieldsMap = Map.set(idx, FieldType::Island(idx), genState.fieldsMap)
}
```

and absolutely no error/text was displayed in the browser nor the console. Just an empty gray rounded rectangle on red background appeared. The issue is lack of space between `{` and `genState`.

gdotdesign:
This happens when either the base `SyntaxError` or `TypeError` class is raised, I'll try to track it down.

Namek:
Another, similar one:

```
{num >= genState.chanceToPickExistingIsland, rand1}
```

The first tuple item is supposed to be a boolean expression. Auto-formatter removes the space after `{` and I get the same empty error.

Gets fixed when I move expression to a variable.

```
shouldPickExistingIsland =
  num >= genState.chanceToPickExistingIsland

{shouldPickExistingIsland, rand1}
```

Namek:
Got a new issue: functions defined inside `where` block are not type-checked. It seems to be a regression since it behaves properly on the Sandbox: https://sandbox.mint-lang.com/sandboxes/odStFrMXwJMkLw

```
module Test {
  fun test1(a : Number) : String {
  	""
  } where {
  	test2 = (a : Number) : String {
      0
    }
  }
}
```

The `test2` function should fail compilation.

Namek:
BTW I'm testing the array destructuring branch but report tuples here.

gdotdesign:
@Namek do you have a SSCCE for the empty error?

gdotdesign:
I've resolved the regression in the `array-destructuring` PR since it's ahead.

gdotdesign:
@Namek managed to track the empty error down.

PR #210, opened 2020-02-26
Array Destructuring

### LINKED ISSUE #177: Feature Request: destructure first array element
In functional languages we usually destructure a list for convenient recursive traversal of some tree/graph data.

For simplicity of an example let's just parse a flat array:

```
state someList = [1, 2, 3, 141424]

fun sumNumbers(elements : Array(Number)) : Number {
  traverse(0, elements)
} where {
  traverse = (acc : Number, els : Array(Number)) : Number {
    case (els) {
      [] => acc
      first::rest => traverse(first + acc, rest)
    }
  }
}
```

The important part is:

```
case (els) {
  [] => acc
  first::rest => traverse(first + acc, rest)
}
```

where `els` is destructured to the `first` element and the `rest` of the elements.

# Why?

In the above example we could use a simple `Array.map()` call or `for` syntax. However, coding anything like breadth/deep first search algorithms would benefit from a simpler syntax.

Currently, I am not aware of a simple way to split a list to first/rest parts in a one call.

gdotdesign:
This PR implements array destructuring (will fix #177).  It allows to match the structure of an array and assign specific items in an array to variables.

It can be used in a case statement:
```
case ([]) {
  [] => ""                /* Empty array  */
  ["a"] => ""             /* Array with only one element */
  [a] => ""               /* Array with ony one element */
  [a, b] => ""            /* Array with ony two elements  */
  [a, ...middle, b] => "" /* Array at least with three elements */
  [...head, tail] => ""   /* Array at least two elements */
  [head, ...tail] => ""   /* Array at least two elements */
  [...items] => ""        /* Items would be the array */
  => ""                   /* Fallback as usual */
}
```
------------------

`Ast::Spread`
- [x] Node
- [x] Parser
- [x] Formatter

`Ast::ArrayDestructuring`
- [x] Node
- [x] Parser
- [x] Compiler
- [x] Formatter
- [x] Type Checking


Sija:
Great work @gdotdesign, it looks awesome! 😻
One thing that looks weird to me is `[a, b, c] =>` notation  (outside of `case` that is), I'd expect `[a, b, c] =` instead.

gdotdesign:
@Sija actually it cannot be used outside of the `case` (reverted that in the last commit) because it doesn't make sense, or I am misunderstanding something? :smile:

gdotdesign:
@Sija actually it cannot be used outside of the `case` (reverted that in the last commit) because it doesn't make sense, or I am misunderstanding something? :smile:

gdotdesign:
@Sija It does in but array can only be deconstructed in `case` expressions because otherwise we cannot guarantee that the required elements exist, with tuples we can because we no their length at compile time.

So with array only this use case is supported https://github.com/mint-lang/mint/blob/a20d3d2ea11179afd8e3b8e5373161d112d50a89/spec/compilers/case_with_array_destructuring
With tuples we can do the destructuring in `where`, `try`, `sequence` and `parallel`  as well like this https://github.com/mint-lang/mint/blob/b29db315db2be43b0bcf612f0939aea44cd57cfa/spec/compilers/where_with_tuple_destructuring

PR #215, opened 2020-03-13
Typo: prdicate => predicate

Tuxified:
Mint looks interesting, but when I started looking at the docs, there seems to be a typo in the first part of the docs?


gdotdesign:
Yep it was a typo, thanks :+1:

Issue #191, opened 2020-02-16
Feature Request: default constructor for record


Namek:
It is easy to construct an instance of a single record, however with multiple records it gets very repetetive and adds no value to readability by actually decreasing it.

For readability and convenience it would be nice to  have a default constructor for any record.

## Current solution

For a record defined as:

```
record Bridge {
  connectionCount : Number,
  orientation : Orientation,
  idx1 : Number,
  idx2 : Number
}
```

I would like to construct a list of record instances:

```
bridges =
  [
    brg(2, Orientation::Horizontal, 0, 3),
    brg(1, Orientation::Vertical, 0, 16),
    brg(1, Orientation::Vertical, 3, 11),
    brg(1, Orientation::Horizontal, 9, 11),
    brg(1, Orientation::Horizontal, 11, 19),
    brg(1, Orientation::Horizontal, 16, 19)
  ]
```

However, I have to manually write construction functions like this:
```
  fun brg (
    connectionCount : Number,
    orientation : Orientation,
    idx1 : Number,
    idx2 : Number
  ) : Bridge {
    {
      connectionCount = connectionCount,
      orientation = orientation,
      idx1 = idx1,
      idx2 = idx2
    }
  }
```

## Proposal

I **propose** to simply **make a function** from a record. For the case above:

```
Bridge(1, Orientation::Horizontal, 16, 19)
```

Of course constructors won't be used for every created Record so it would be consistent to analyze usage for **dead code elemination**.

Namek:
Additional advantage of having record constructors would be Tuples. We don't have Tuples at this moment, neither in syntax nor standard library. This would be a first step. See #193

gdotdesign:
I'm not against having constructors per say, I just like defining things explicitly, however I see your point about records in arrays, is there anything besides that which would be easier with  constructors? 

Maybe we could figure out something else which keeps the explicitness a record (naming wise) but would make it easier to construct an array of records?

Namek:
> is there anything besides that which would be easier with constructors?

Notice the `inputPos`:
```
fun onPointerMove (evt : Html.Event) : Promise(Never, Void) {
  try {
    inputPos =
      {
        x = evt.offsetX,
        y = evt.offsetY
      }

    next { }
   }
}
```

The type of `inputPos` is not explicitly defined. In this case I could do `Vec2(evt.offsetX, evt.offsetY)`. (where `Vec` stands for `Vector`)

[...]

Issue #132, opened 2019-06-13
Required properties


gaku-sei:
Required properties would allow the component to leave the property uninitialized!

```
required property value : String /* No default */
```

This would probably lead to some error at compile time if the parent doesn't provide any values.

gdotdesign:
This can be implement but I'm not sure it's necessary or not :slightly_smiling_face: 

I never needed to leave out the default value and besides it's nice documentation on what the actual value should be.

gdotdesign:
Closing since I answered in the previous comment and no additional discussion took place.

Feel free to open it again if you would like to discuss it further.

luncheon:
I think a compilation error indicating `the required property is not provided` can help us fix the code.

How do you think?

gdotdesign:
I still don't feel that we need this feature.

:+1: if you would like to have required properties (without a default value) :-1: if you don't

Namek:
@luncheon @gaku-sei can you provide us some cases?

luncheon:
* url of a link or image
* label of a tab
* complex record property: providing default value is boring and using `Maybe` makes code dirty

luncheon:
```
record User {
  id : String
  name : String
  email : String
  /* many properties */
}

component UserTable {
  property users : Array(User) = []

  fun render : Html {
    <table>
      for (user of users) {
        <UserRow user={user}/>
      }
    </table>
  }
}

component UserRow {
  property user : User /* required */
}
```

gdotdesign:
@gaku-sei thanks for the use-cases the record one is one that I didn't think about.

PR #216, opened 2020-03-14
🚧 Type System Improvements

gdotdesign:
This PR is an RFC for visual improvements for the code itself. 

I feel like that Mint code has some visual clutter which can be resolved quite easily, but for some  we need type inference.

Removing some elements helps reduce visual complexity, makes it easier to read and reason about the code, also it has the additional benefit of writing less.

### Types

For **functions**, **states**, **properties** and **computed properties** the type (or return type) is required, and for me sometimes it's a burden to write them. **These type definitions can be made optional.**

The only downside of that is that for recursive functions and for the component records we need to calculate **static type signatures** and without a defined type we can use a type variable `a` but having a type for helps with that.

Type definitions elsewhere will still be required (arguments, provider, records, etc...)

### The `fun` keyword

It is possible to remove the `fun` keyword which in hindsight is not needed.

----------------------

Here is an example of how a component file looks now and how it can look with this PR:

```
global component LoadingOverlay {
  state loading : Bool = false

  style base {
    cursor: progress;
    position: fixed;
    z-index: 10000;
    bottom: 0;
    right: 0;
    left: 0;
    top: 0;
  }

  fun show : Promise(Never, Void) {
    next { loading = true }
  }

  fun hide : Promise(Never, Void) {
    next { loading = false }
  }

  fun render : Html {
    if (loading) {
      <div::base/>
    } else {
      <></>
    }
  }
}
```

```
global component LoadingOverlay {
  state loading = false

  style base {
    cursor: progress;
    position: fixed;
    z-index: 10000;
    bottom: 0;
    right: 0;
    left: 0;
    top: 0;
  }

  show {
    next { loading = true }
  }

  hide {
    next { loading = false }
  }

  render {
    if (loading) {
      <div::base/>
    } else {
      <></>
    }
  }
}
```

---------------------

This PR can be merged as-is, all the tests pass, however I still plan to work on this. 

The `fun` keyword is optional but the compiler throws a warning that it will be removed in the future and the formatter automatically removes them.

Please give your feedback and if there is anything else which you think can be removed just let me know.


s0kil:
The `Promise(Never, Void)` Type is very common, this PR will make it easier.

Namek:
I would leave `fun` as it is. It's easier to jump in view of I can search a function definition when it is prefixed.

Other than that, I would like to have an ability to restrict/write types of variables in `try` block.

gdotdesign:
> Other than that, I would like to have an ability to restrict/write types of variables in try block.

It's easy to add it as an optional thing, I can see that it could be helpful from some situations, I never needed myself it though.

> I would leave fun as it is. It's easier to jump in view of I can search a function definition when it is prefixed.

I have mixed feelings here as well. GitHub need to implement some kind of polls for issues :joy:

Sija:
@gdotdesign It's pretty easy!

Add a 👍 reaction to this comment for leaving `fun` or 👎 for removing it :)

spec/formatters/connect
```diff
@@ -6,7 +6,7 @@ store Test {
 }
 
 component A {
-  connect Test exposing{a,b,c,d}
+  connectTest exposing{a,b,c,d}
```
Namek: this doesnt seem right

spec/type_checking/function_call
```diff
@@ -80,35 +80,19 @@ component Main {
 --------------------------------------------------------------------------------
 module Maybe {
   fun nothing : Maybe(a) {
-    `new Nothing()`
+    ``
   }
 
   fun just (value : a) : Maybe(a) {
-    `new Just(value)`
+    ``
```
Namek: why is it empty now?

spec/formatters/connect
gdotdesign: Strictly speaking the parser does not need white-space between the keyword and the name of the store here. 

It's because there is no tokeniser phase, the parser directly produces the AST.

spec/type_checking/function_call
gdotdesign: These tests does not execute JavaScript code, it just parses the code and type checks based on the annotations, so it does not make sense having that here.

gdotdesign:
By popular demand the `fun` keyword will not be removed :slightly_smiling_face: and changes were reverted.

Now this PR contains the following:
- optional return type definition for **functions** and **inline functions**
- optional type definition for **state**, **get** and **property**
- default value for properties can now be omitted #132
- restricting type of arrays: `[] of Number`
- restricting type of inline JavaScripts `\`"Hello"\` as String`

src/type_checker.cr
```diff
@@ -55,6 +54,31 @@ module Mint
...
+              i.name.value
+            when Ast::With
+              "<with>"
+            when Ast::Try
+              "<try>"
+            when Ast::Call
+              "<call>"
+            else
+              i
+            end
```
Sija:
```suggestion
        x = case i
            when Ast::Component then i.name
            when Ast::Function  then i.name.value
            when Ast::With      then "<with>"
            when Ast::Try       then "<try>"
            when Ast::Call      then "<call>"
            else
              i
            end
```

gdotdesign:
Two more issues has been implemented here:
- Pipe operator overhaul (can be piped into expressions instead of just calls) #228 
- Record constructors #191

Issue #221, opened 2020-03-19
Avoiding unknown CSS property error


luncheon:
```
There is no CSS property with the name: touch-action
```

I think it is too hard to support all CSS properties.  
Is there a way to avoid the error? If not, is there a plan to provide it?

I considered two ways:

1. syntax to avoid checking CSS properties
2. config to add allowed CSS properties

and I think 2 is safer.

s0kil:
We should support all CSS properties and values that browser engines support.
## Chromium:
- [Primitive Values](https://github.com/chromium/chromium/blob/master/third_party/blink/renderer/core/css/css_primitive_value_units.json5)
- [Properties](https://github.com/chromium/chromium/blob/master/third_party/blink/renderer/core/css/css_properties.json5)
- [Values](https://github.com/chromium/chromium/blob/master/third_party/blink/renderer/core/css/css_value_keywords.json5)
- [Media Types](https://github.com/chromium/chromium/blob/master/third_party/blink/renderer/core/css/media_type_names.json5)
- [Media Features](https://github.com/chromium/chromium/blob/master/third_party/blink/renderer/core/css/media_feature_names.json5)
- [SVG Values](https://github.com/chromium/chromium/blob/master/third_party/blink/renderer/core/css/svg_css_value_keywords.json5)

s0kil:
@luncheon Please re-open issue, I'm saying we should support all CSS property/values that browser engines support, but we currently do not.

Example, [`touch-action`](https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action) is a valid property, but it's not [defined in Mint](https://github.com/mint-lang/mint/blob/master/src/assets/css_properties)

s0kil:
@gdotdesign Should we also define properties with a prefix such as `-webkit-`?
Otherwise, we could figure out that `transform` could also be `-webkit-transform`.

s0kil:
@gdotdesign In the mentioned "properties-db.js" file, it also defines the available values for a property, we could use that to check the values?

gdotdesign:
that is something I want to do and that file seems to be prefect for it :slightly_smiling_face: it's just not a priority right now so if you want to tackle it it would be awesome :smile:

PR #222, opened 2020-03-19
Added more CSS properties

s0kil:
Added more CSS properties according to https://github.com/chromium/chromium/blob/master/third_party/blink/renderer/core/css/css_properties_ranking.json5

Addresses the compiler error at https://github.com/mint-lang/mint/issues/221


src/assets/css_properties
```diff
@@ -1,3 +1,12 @@
+-epub-caption-side
+-epub-text-combine
+-epub-text-emphasis
+-epub-text-emphasis-color
+-epub-text-emphasis-style
+-epub-text-orientation
+-epub-text-transform
+-epub-word-break
+-epub-writing-mode
```
s0kil: Is this valid in Mint?

gdotdesign: It is, and properties starting with `-` (mostly browser prefixed) are not checked so we don't need to add these.

s0kil: Could we also check those? What other properties are there? Besides `-webkit-`, `-moz-`, `-ms-`, `-o-`?

gdotdesign: https://github.com/mint-lang/mint/blob/master/src/type_checkers/css_definition.cr#L21

gdotdesign: We could check those but there are a lot of them and keeping those up to date would be a challenge.

Also they are not used very much so I don't think it worth the effort.

s0kil: Looking at Autoprefixer, the mentioned prefixes are the only ones they use.
But likes you said, now days it is automated.

s0kil: Looks good to merge!

src/assets/css_properties
```diff
@@ -552,4 +653,7 @@ wrap-flow
 wrap-inside
 wrap-through
 writing-mode
+x
+y
 z-index
+zoom
```
Sija: Missing newline.

s0kil: Is it going to break?

Sija: I guess not, but it's a good practice to have files with trailing newline.

gdotdesign: I don't think so, it's just code style: I have it setup that it always adds a new line on save if there is not one.

Sija: @gdotdesign It's more than that, see https://thoughtbot.com/blog/no-newline-at-end-of-file for more in-depth explanation.

s0kil: Dam it Unix.

Issue #230, opened 2020-03-24
Issues starting mint server


petemcfarlane:
Hi, I have a couple of small bug bears when running `mint start` development server. The first I think I can submit a PR for, the second I'm not experienced enough with Crystal language to see why it's not working!

Firstly, if default port 3000 is busy, it asks to try a different port and asks if I'd like to select it (Y/n). When I see a capitol "Y" that _normally_ means I can just hit return and that will be the default option. 
But as you can see in the gif below that is not the case here.
![Screen Recording 2020-03-24 at 11 41 am](https://user-images.githubusercontent.com/3472717/77422328-2bdb5200-6dc5-11ea-8572-9d1d12353050.gif)

Also, it asks if I want to start on 3001, and it does indeed start on 3001, but you can see that the cli output still says it started on port 3000 (which it didn't) - I can see the commit where this broke here: https://github.com/mint-lang/mint/commit/12b71f8290c5e368b16751bf96160606002e49c2#diff-361e218ff8acd300b1ce3017145bf132R31 but not sure how to best fix it, without all the context, sorry!

PR #231, opened 2020-03-24
Default to true, for use_new_port

petemcfarlane:
When running `mint start` if default port is taken, allow return (with no explicit Y) to start dev server.
Partly solves #230

### DIFF
--- a/src/utils/server.cr
+++ b/src/utils/server.cr
@@ -38,7 +38,7 @@ module Mint
         terminal.print "#{COG} Would you like to to use port #{new_port} instead? (Y/n)\n"
 
         use_new_port = gets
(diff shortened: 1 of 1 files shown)


src/utils/server.cr
```diff
@@ -38,7 +38,7 @@ module Mint
         terminal.print "#{COG} Would you like to to use port #{new_port} instead? (Y/n)\n"
 
         use_new_port = gets
-        if !use_new_port.nil? && use_new_port.downcase == "y"
+        if !use_new_port.nil? && use_new_port == "" || use_new_port.downcase == "y"
```
Sija:
```suggestion
        if !use_new_port.nil? && use_new_port.empty? || use_new_port.downcase == "y"
```

gdotdesign:
Thanks :+1:

PR #236, opened 2020-03-29
Do not override `Object#to_s`

veelenga:
Following up the convo here https://github.com/mint-lang/mint/commit/364af52f3d7bf139f008592db3ffd49859ec0480#r38111871

The problem is `Object#to_s` is overridden which is not desired in most cases. In this PR we change it to`Object#to_s(io)` instead, which is used by `Object#to_s` internally. 

<hr>

Check out the warning here:
 https://github.com/crystal-lang/crystal/blob/612825a53c831ce7d17368c8211342b199ca02ff/src/object.cr#L89-L93

Check out the `Array#to_s(io)` sample:
https://github.com/crystal-lang/crystal/blob/612825a53c831ce7d17368c8211342b199ca02ff/src/array.cr#L1825

### DIFF
--- a/.ameba.yml
+++ b/.ameba.yml
@@ -1,4 +1,2 @@
 Metrics/CyclomaticComplexity:
   Enabled: false
-Lint/RedundantStringCoercion:
-  Enabled: false
--- a/src/ast/data.cr
+++ b/src/ast/data.cr
@@ -6,13 +6,8 @@ module Mint
       def initialize(@input : String, @file : String)
       end
 
-      def to_s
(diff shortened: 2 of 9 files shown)


gdotdesign:
It's already addressed in #235 by @Sija with the difference is that it is more performance aware based on https://crystal-lang.org/reference/guides/performance.html#dont-create-intermediate-strings-when-writing-to-an-io

So I'll probably use that, but thanks for taking the time and creating this PR!

veelenga:
Sorry, I haven't noticed it. Thanks!

PR #237, opened 2020-03-29
Several more refactors & optimizations

Sija:
Followup to #235

### DIFF
--- a/src/render/terminal.cr
+++ b/src/render/terminal.cr
@@ -10,12 +10,16 @@ module Mint
         end
 
         def close
-          print "\n" unless @last =~ /\n|\r/
+          print "\n" unless @last.in?("\n", "\r")
+        end
+
+        def print(contents : String)
+          @last = contents.last
+          @io.print contents
         end
 
         def print(contents)
-          @last = contents.last
-          @io.print contents
+          print contents.to_s
         end
 
         def text(contents)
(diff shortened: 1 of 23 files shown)


gdotdesign:
Nice! You are on fire today :fire: :+1:

src/render/terminal.cr
```diff
@@ -55,29 +64,29 @@ module Mint
...
               @cursor = 0
               print char
-            when .=~(/\s/)
-              @cursor += char.size
+            when .whitespace?
+              @cursor += 1
               print char
             else
-              @cursor += char.size
+              @cursor += 1
```
Namek: What about 2-byte characters? UTF-16

Sija: This was done in order to keep the logic from previous implementation.

gdotdesign: I think it should be size since it returns the number of Unicode codepoints https://crystal-lang.org/api/0.33.0/String.html#size-instance-method

Sija: @gdotdesign _Unicode codepoints_ in this case is equals to a character - see closely the included example code in the docs - it shows you the amount of _characters_, not bytes.

also, it's not a `String` anymore, it's a `Char`, so this method doesn't apply here.

gdotdesign: All right, I have run into some multiple unicode code point characters at some point but I can't remember what it was so probably not that important.

And I also vaguely remember that I needed to cast the character to a string for some reason, but if this works correctly I'm all for it :slightly_smiling_face:

Sija:
@gdotdesign btw, I've noticed one inconsistency:

https://github.com/mint-lang/mint/blob/218d27872dd2cbaf8adf53d906cce57dd1521483/src/test_runner.cr#L300-L334

Some calls to `puts` are invoked with `terminal` as a receiver and some don't - is it intentional behaviour?

gdotdesign:
> Some calls to `puts` are invoked with `terminal` as a receiver and some don't - is it intentional behaviour?

It was not intentional :sweat_smile:

gdotdesign:
Thanks :tada:

PR #253, opened 2020-04-25
Fix #252

### LINKED ISSUE #252: Not enough context in required props error
I got this error while working on my site:

image

There's a component somewhere that isn't getting a required prop passed to it. But there's no indication of _where_ that is or what component.

s0kil:
In the [`ComponentMainProperty`](https://github.com/mint-lang/mint/blob/master/src/messages/component_main_property.cr) and [`HtmlComponentAttributeRequired`](https://github.com/mint-lang/mint/blob/master/src/messages/html_component_attribute_required.cr) the `property` in the error message does not show, not sure why.

### DIFF
--- a/src/messages/component_main_property.cr
+++ b/src/messages/component_main_property.cr
@@ -7,6 +7,6 @@ message ComponentMainProperty do
     text "component cannot have properties."
   end
 
-  snippet property_name, "A property is defined here:"
+  snippet property_node, "A property is defined here:"
   snippet node, "The component is here:"
 end
--- a/src/messages/component_main_property.cr
+++ b/src/messages/component_main_property.cr
@@ -7,6 +7,6 @@ message ComponentMainProperty do
     text "component cannot have properties."
   end
 
-  snippet property, "A property is defined here:"
+  snippet property_name, "A property is defined here:"
   snippet node, "The component is here:"
 end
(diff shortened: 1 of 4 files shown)


Sija:
> Could we take advantage of Crystal's type checker is such cases?

@s0kil We could add type restrictions to the method arguments to avoid such situations, I guess.

s0kil:
Crystal actually shows warning saying that [this](https://github.com/mint-lang/mint/blob/master/src/message.cr#L36-L44) case expression is not exhaustive.

@Sija Shouldn't that be an error instead?

Sija:
@s0kil Good catch! In the next Crystal version it will be by default.

Sija:
Let's leave it for another PR fixing all of those cases.

s0kil:
> I'm not sure about the rename of the property node variable, but other than that LGTM

To be honest, I could not figure out the root of the problem, but the name `property` is interfering and somehow becomes Nil.
Maybe you would have a better idea, is this something specific to Crystal?

s0kil:
@Sija Maybe this is why? https://github.com/crystal-lang/crystal/wiki/Crystal-for-Rubyists#properties

s0kil:
@gdotdesign Could you comment on the behavior of `property`?

zephraph:
This is so much better! Thanks for the quick turn around:

image

gdotdesign:
@s0kil without doing some investigation I think the `property` here is considered a macro https://crystal-lang.org/api/0.34.0/Object.html#property(*names,&block)-macro

PR #255, opened 2020-04-26
Tweak Travis CI setup

Sija:


zephraph:
Hmmm, why was `make development` removed here? If that's intentional, should we update the contributing docs?

gdotdesign:
I didn't really use the `development` task just the `build` task, but it can be useful.

PR #256, opened 2020-04-27
Followup to #255

Sija:
Adds back `skip_cleanup: true` flag to deploy steps on Travis CI, since apparently we’re still using deployments v1, in contrast to the _hint_ displayed within the Travis' web ui.

Followup to #255

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ before_deploy:
 
 deploy:
   - provider: s3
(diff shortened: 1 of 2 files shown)


s0kil:
@Sija Could we check if `~/.bin/` or `~/.local/bin` exists otherwise create the directory before the `mint` binary is copied? Making it a tad more convenient.

PR #257, opened 2020-04-27
Reorganized README

s0kil:
Moved the badges to the footer, we should get away from the popular choice, the badges add unnecessary clutter when a person visits the README.
Also, added Mint logo to the top, and the survey link.


gdotdesign:
I like badges at the top but I'm fine with if they are at the bottom. Can you show some other repos where they are at the bottom?

s0kil:
> I like badges at the top but I'm fine with if they are at the bottom. Can you show some other repos where they are at the bottom?

I do not know of any, the goal was to reduce cognitive overload when visiting the repo, for example: focus on reading about mint without the distracting colors coming from the badges.

PR #258, opened 2020-04-27
Refactors round 3

Sija:


src/compilers/html_element.cr
```diff
@@ -38,8 +38,8 @@ module Mint
       attributes =
         node
           .attributes
-          .reject(&.name.value.==("class"))
-          .reject(&.name.value.==("style"))
+          .dup
```
Namek, line 41: why?

src/render/terminal.cr
```diff
@@ -82,7 +82,7 @@ module Mint
             when '\n', '\r'
               @cursor = 0
               print char
-            when .whitespace?
+            when .ascii_whitespace?
```
Namek, line 85: can you explain? what's the impact?

src/style_builder.cr
```diff
@@ -192,17 +192,17 @@ module Mint
...
     def process(node : Ast::CssSelector,
-                id : String | Nil,
-                at : String | Nil,
+                id : String?,
+                at : String?,
                 parents : Array(String),
                 conditions : Array(String),
                 style_node : Ast::Node)
-      selectors = [] of String
+      selectors = %w[]
```
Namek, line 205: just an opinion here but `[] of String` is more readable to newcomers

src/type_checkers/types.cr
```diff
@@ -219,25 +219,26 @@ module Mint
         node1 = prune(node1)
         node2 = prune(node2)
 
-        if node1.is_a?(Variable)
-          if node1 != node2
+        case
+        when node1.is_a?(Variable)
+          unless node1 == node2
```
Namek, line 224: just an opinion but `unless` reduces understandability. The first thing I see is `==`. And then I have to invert it because there is some `unless`. I wouldn't write that. I know people write like that a lot in Ruby community but there are more communities in the world and I rarely see this.

src/compilers/html_element.cr
Sija: Because of `reject!` one line below.

src/render/terminal.cr
Sija: `Char#whitespace?` includes also unicode characters, which in most of the cases we won't care about.

src/style_builder.cr
Sija: Perhaps, but that's what the language reference is for, I don't see a point in using only newcomers-friendly syntax, it ain't a language playground.

src/type_checkers/types.cr
Sija: I tend to read it from left to right, so `unless` comes first, so that reads more naturally for me, also I like the positive affirmations (`==`) vs the negative (`!=`). And yes, it comes in part from Ruby, but Crystal **is** a Ruby-based (syntax-wise) language after all.

src/render/terminal.cr
Namek: So we can't have emoji? Or we're closing to Chinese community? Because of micro optimization?

gdotdesign: It's just about whitespace right? Unicode whitespace characters specifically. Also this is just a method which prints stuff the console in a specific width.

src/mint_json.cr
```diff
@@ -14,34 +14,42 @@ module Mint
...
     json_error MintJsonRootInvalidKey
 
-    getter test_directories, source_directories, dependencies, application
-    getter external_files, name, root, formatter_config
-
     def self.parse_current : MintJson
       path = File.join(Dir.current, "mint.json")
       new File.read(path), Dir.current, path
-    rescue exception : IO::Error
+    rescue ex : IO::Error
```
gdotdesign, line 37: What is wrong with `exception` ? :smile:

src/parsers/number_literal.cr
```diff
@@ -23,7 +23,7 @@ module Mint
 
         Ast::NumberLiteral.new(
           from: start_position,
-          value: BigDecimal.new(value),
+          value: value.to_big_d,
```
gdotdesign, line 26: This method name is gold :joy:

src/parsers/parallel.cr
```diff
@@ -17,7 +17,7 @@ module Mint
           results = many { statement(Ast::Statement::Parent::Sequence) || comment }.compact
 
           raise ParallelExpectedStatement if results
-                                               .reject(&.is_a?(Ast::Comment))
+                                               .reject(Ast::Comment)
```
gdotdesign, line 20: `reject!`?

Sija: I don't think we want to modify the receiver in this case.

src/mint_json.cr
Sija: Oops 😅

src/parsers/parallel.cr
gdotdesign: yes, you are right, sorry.

src/render/terminal.cr
Sija: @gdotdesign Exactly! @Namek seemed to missed the point about what are we talking about here 🤣

PR #271, opened 2020-05-03
add entries to Map

ymmooot:
Hi mint-lang 🍃.
Map has `keys` and `values` but no `entires` so I added.
It's very useful to iterate maps.

### DIFF
--- a/core/source/Map.mint
+++ b/core/source/Map.mint
@@ -314,4 +314,26 @@ module Map {
   fun size (map : Map(key, value)) : Number {
     `#{map}.size`
   }
+
+  /*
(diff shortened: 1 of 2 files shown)


s0kil:
@ymmooot Thank You 🍃

gdotdesign:
@ymmooot thanks :)

Probably you don't know but you can iterate on maps using the for loop as well:

```
for (key, value of map) {
  ...
} 
```

Here is an example: https://sandbox.mint-lang.com/sandboxes/9C6pcDPdUmWzLA

---------

And now checking it it's not documented :sweat_smile:

PR #283, opened 2020-05-06
Feature/inner destructuring

teggotic:


Sija:
@teggotic Ameba warnings need to be fixed, see failed CI builds.

src/ast.cr
```diff
@@ -4,6 +4,8 @@ module Mint
 
     alias TypeOrVariable = Type | TypeVariable
 
+    alias DestructuringType = ArrayDestructuring | EnumDestructuring | TupleDestructuring
```
gdotdesign, line 7: This is not necessary and should be replaced with `Node`

src/ast/enum_destructuring.cr
```diff
@@ -1,9 +1,11 @@
 module Mint
   class Ast
     class EnumDestructuring < Node
+      alias EnumDestructuringParameter = DestructuringType | TypeVariable
```
gdotdesign, line 4: Also not necessary.

src/compilers/array_destructuring.cr
```diff
@@ -1,36 +1,64 @@
...
-          .items
-          .reverse
-          .take_while(&.is_a?(Ast::Variable))
-          .each do |var|
-            statements << "const #{js.variable_of(var)} = __.pop()"
-          end
-
-        statements << "const #{js.variable_of(node.items.select(Ast::Spread).first.variable)} = __"
-        statements
+        condition = "Array.isArray(#{condition_var_name}) && #{condition_var_name}.length >== #{node.items.size - 1}"
```
gdotdesign, line 23: This can be `condition = if` makes it easier to read.

src/compilers/array_destructuring.cr
```diff
@@ -1,36 +1,64 @@
 module Mint
   class Compiler
-    def _compile(node : Ast::ArrayDestructuring, value)
+    private macro compile_item!(initial_var, item = item, index = index1)
```
gdotdesign, line 3: Why does this needs to be a macro?

src/parsers/array_destructuring.cr
```diff
@@ -31,5 +31,8 @@ module Mint
           input: data)
       end
     end
+
+    def parse_destructuring_item
```
gdotdesign, line 35: Left over empty method.

src/parsers/enum_destructuring.cr
```diff
@@ -11,7 +12,16 @@ module Mint
 
         option = type_id! EnumDestructuringExpectedOption
 
-        parameters = many { type_variable }.compact
+        parameters = (many do
```
gdotdesign, line 15: I think we should have only one syntax, in this case with parentheses. This will be a breaking change, but would fix the else branch since there it's not possible to inner destructure.

src/type_checkers/case_branch.cr
```diff
@@ -45,19 +45,9 @@ module Mint
...
-              when Ast::Spread
-                {variable.variable.value, condition, variable.variable}
-              else
-                # ignore
-              end
-            end
+          variables = get_destructuring_variables item, condition
 
-          scope(variables) do
+          scope(variables.not_nil!) do
```
gdotdesign, line 50: I suggest `to_a` instead of `not_nil!`

src/type_checkers/enum_destructuring.cr
```diff
@@ -33,6 +33,12 @@ module Mint
           "option" => option,
           "node"   => param,
         } unless option.parameters[index]?
+
+        case param
+        when Ast::EnumDestructuring
```
gdotdesign, line 38: Other branches should be resolved as well, since they can cause other errors.

src/parsers/enum_destructuring.cr
teggotic: Maybe we want to have a comma between parameters?

Sija:
@teggotic Hi, are you willing to continue working on this?

teggotic:
> @teggotic Hi, are you willing to continue working on this?

You can take it

gdotdesign:
Superseded by #593

Issue #276, opened 2020-05-03
RFC: Add Websocket to core library


lukepighetti:
I found this example of a web socket but it would be nice if it was included in the core library. 

https://sandbox.mint-lang.com/sandboxes/N0iQTG1fduXoRA

Somewhat related discussion: are platform channel implementations (back tick JS) allowed in the core library? I admit I'm not too familiar with `Crystal` or the `mint` language architecture in general.

gdotdesign:
Yes it's my intention to move this into the core library, it just needs some clean up and some tests.

> are platform channel implementations (back tick JS) allowed in the core library?

Most of the core library is build using inline JavaScript (``` `` ```) if that's what you mean. The core library is in this repo as well https://github.com/mint-lang/mint/tree/master/core

gdotdesign:
It's added with the #286 PR.

PR #286, opened 2020-05-18
Improve providers.

### LINKED ISSUE #285: Websocket Provider does not update when Subscription changes
see:  https://sandbox.mint-lang.com/sandboxes/N0iQTG1fduXoRA

gdotdesign:
This PR is for improving providers, but most importantly to be able to write provides without inline JavaScript:
- now they can have `state` like stores and components
- migrated all providers from inline JavaScript to Mint code
- added helper functions from native code (event listeners, animation frame)
- the `attach` and `detach` functions are replaced with `update`
- the `update` function is called if the subscription data changes (will fix #285)
- added `Websocket` module and provider #276 
- added `IntersectionObserver` module and provider


core/source/AnimationFrame.mint
```diff
@@ -0,0 +1,20 @@
...
+  fun request (method : Function(a)) : Number {
+    `requestAnimationFrame(#{method})`
+  }
+
+  /*
+  Cancels a scheduled function call.
+
+    AnimationFrame.cancel(id)
+  */
+  fun clear (id : Number) : Number {
```
Sija, line 17:
```suggestion
  fun cancel (id : Number) : Number {
```

core/source/WebSocket.mint
```diff
@@ -0,0 +1,106 @@
+record WebSocket.Config {
+  onOpen : Function(WebSocket, Promise(Never, Void)),
+  onMessage : Function(String, Promise(Never, Void)),
+  onError : Function(Promise(Never, Void)),
+  onClose : Function(Promise(Never, Void)),
+  reconnect : Bool,
```
Sija, line 6:
```suggestion
  reconnectOnClose : Bool,
```

core/source/AnimationFrame.mint
Sija, line 17:
```suggestion
  fun cancel (id : Number) : Number {
```

PR #287, opened 2020-05-20
Convert Map and Set to use arrays to be able to support any type.

### LINKED ISSUE #268: Cannot use Enum's as a key for Map
Hey, 

Not sure if I'm just using it wrong, but it looks like you cannot use enum members as keys in a Map.

Here's an example of this in action:
https://sandbox.mint-lang.com/sandboxes/YztcefM7LQvBSA
```
enum TestEnum {
  Field1
  Field2
  Field3
  Field4
}
component Main {
  property map : Map(TestEnum, String) =
    Map.empty()
    |> Map.set(TestEnum::Field1, "Value 1")
    |> Map.set(TestEnum::Field2, "Value 2")
    |> Map.set(TestEnum::Field3, "Value 3")

  fun getMapValue (test : TestEnum) : String {
    Debug.log(map)
    |> Map.get(test)
    |> Maybe.withDefault("Unknown")
  }

  fun render : Html {
    <div>
      <{ getMapValue(TestEnum::Field1) }>
      <{ getMapValue(TestEnum::Field2) }>
      <{ getMapValue(TestEnum::Field3) }>
      <{ getMapValue(TestEnum::Field4) }>
    </div>
  }
}
```
This should print "Value 1Value 2..." etc but instead prints "Unknown" 4 times.

gdotdesign:
Resolves #268

### DIFF
--- a/core/source/Map.mint
+++ b/core/source/Map.mint
@@ -2,7 +2,7 @@
 module Map {
   /* Returns an empty map. */
   fun empty : Map(x, z) {
-    `new Map()`
+    `[]`
   }
 
   /*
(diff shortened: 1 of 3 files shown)


core/source/Map.mint
```diff
@@ -77,21 +87,10 @@ module Map {
...
-
-      for (let item of #{map2}) {
-        map.set(item[0], item[1])
-      }
-
-      return map
-    })()
-    `
+    map2
+    |> Map.reduce(map1, (memo : Map(x, z), key : x, value :z) {
```
Sija, line 91:
```suggestion
    |> Map.reduce(map1, (memo : Map(x, z), key : x, value : z) {
```

PR #289, opened 2020-06-04
Template 'left' and 'top' in `Dom.getElementFromPoint`

pvdvreede:
This fixes a bug where using the `getElementFromPoint` would give a 'left is not defined' error in JS as the variable was hard coded rather than templated.


gdotdesign:
Thanks :+1:

PR #292, opened 2020-06-25
Changes For Crystal 0.35.x

s0kil:
We should probably disable `--error-on-warnings` flag, not sure if CI will pass.


gdotdesign:
Thanks!

PR #293, opened 2020-06-25
Fix documentation typos

Vechro:
There's one commit spread across 19 files or so because I didn't feel it would be a great idea to make individual commits for a few single-letter fixes per file. Array.mint had quite a few typos so it got a commit of its own.

**Breaking changes**

* String.isAnagarm is now renamed to String.isAnagram
* Provider.Shortcuts now returns the correct keycode for pressing shift

### DIFF
--- a/core/source/Html.mint
+++ b/core/source/Html.mint
@@ -1,6 +1,6 @@
 module Html {
   /*
-  Returns an empty Html node. It is useful for example if you dont to
+  Returns an empty Html node. It is useful for example if you don't to
   render something conditionally.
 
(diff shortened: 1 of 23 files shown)


core/source/Html.mint
```diff
@@ -1,6 +1,6 @@
 module Html {
   /*
-  Returns an empty Html node. It is useful for example if you dont to
+  Returns an empty Html node. It is useful for example if you don't to
```
Sija, line 3: `don't` -> `don't want`

PR #321, opened 2020-09-16
Use Relative Asset Path If --relative Flag Is Set

### LINKED ISSUE #320: Set `manifest.json` as relative path when `--relative` flag is set
https://github.com/mint-lang/mint/blob/master/src/utils/index_html.cr#L32

s0kil:
Fixes #320


gdotdesign:
Great :+1: Thanks!

Issue #278, opened 2020-05-04
RFC: mint format that takes text as arguments


lukepighetti:
The way we currently handle formatting in `mint-vscode` is to copy a file to `tmp`, run `mint format /tmp/myfile.mint`, then read the file and pass it to VSCode.

I'm requesting a feature that allows us to pass a string to `mint format` and have it respond with the formatted result as a string.

This is honestly a minor improvement over what we have now, but it should reduce the format command response time from ~600ms to ~50ms and reduce the chance of collisions in naive implementations.

Any thoughts?

s0kil:
I have thought about this yesterday, and the conclusion is that the language server will also be responsible for formatting code.

We will not need to call `mint format` directly.

gdotdesign:
It should not be a big feature and it comes handy some times (formatting the sandbox code for example).

We should use a flag like `--stdin` to specify inline formatting.

PR #323, opened 2020-09-18
Format Code From STDIN If `--stdin` Flag Is Passed

s0kil:
Implements and closes https://github.com/mint-lang/mint/issues/278

### DIFF
--- a/src/commands/format.cr
+++ b/src/commands/format.cr
@@ -8,7 +8,32 @@ module Mint
       define_argument pattern,
         description: "The pattern which determines which files to format"
 
+      define_flag stdin : Bool,
+        description: "Formats Mint code from STDIN",
+        default: false
+
       def run
+        if flags.stdin
+          self.format_stdin
+        else
+          self.format_files
+        end
+      end
+
+      private def format_stdin
+        input =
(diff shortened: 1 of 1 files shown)


gdotdesign:
Great! :+1:

src/commands/format.cr
```diff
@@ -8,7 +8,32 @@ module Mint
...
+        input =
+          STDIN.gets_to_end
+
+        artifact =
+          Parser.parse(input, "stdin.mint")
+
+        formatted =
+          Formatter.new(artifact, MintJson.parse_current.formatter_config).format
+
+        puts formatted
```
Sija, line 33: @s0kil This should use `terminal.puts` instead.

src/commands/format.cr
```diff
@@ -8,7 +8,32 @@ module Mint
...
+      define_flag stdin : Bool,
+        description: "Formats Mint code from STDIN",
+        default: false
+
       def run
+        if flags.stdin
+          self.format_stdin
+        else
+          self.format_files
+        end
```
Sija, line 20: Redundant `self.`.

PR #329, opened 2020-10-06
Update HtmlComponentExpectedClosingBracket error message

matthewmcgarvey:
Let me know if there's better wording. I tried to add a Link class to put the link in an anchor tag but I kept getting a missing hash key exception.

CleanShot 2020-10-06 at 00 29 54@2x


gdotdesign:
> Let me know if there's better wording. I tried to add a Link class to put the link in an anchor tag but I kept getting a missing hash key exception.

It's fine thanks :+1: there is no way to post a link yet.

PR #336, opened 2020-11-15
GitHub Actions workflow tweaks

Sija:
- Splits workflows into a multiple jobs instead of single job with multiple steps - adding visibility and parallel job execution
- Adds CI for Crystal nightly
- Runs CI on schedule (every Monday at 6am)
- Fixed some things left before having full Crystal 1.0 support
- Misc. low-key tweaks

### DIFF
--- a/.github/workflows/linux-ci-nightly.yml
+++ b/.github/workflows/linux-ci-nightly.yml
@@ -1,6 +1,13 @@
 name: Linux CI (nightly)
 
-on: [push, pull_request]
+on:
+  push:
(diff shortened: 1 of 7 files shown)


gdotdesign (changes requested):

gdotdesign:
Thanks! I didn't know that tests can run in parallel :sweat_smile: 

One thing that could be better if would be a way to only have one `shards install` command and the steps using it.

gdotdesign (approved):

Issue #339, opened 2020-11-28
Confusing error in NextCallStateTypeMismatch


matthewmcgarvey:
<img width="558" alt="CleanShot 2020-11-28 at 12 06 29@2x" src="https://user-images.githubusercontent.com/17329408/100522954-37f88a00-3172-11eb-9221-1132ddff396e.png">

I get that the error is because the value being assigned to state doesn't match the type but the error message doesn't make any sense.

https://github.com/mint-lang/mint/blob/f1b79d086e532b31433b9745145f1015c8db23f5/src/messages/next_call_state_type_mismatch.cr#L4-L9

Sija:
@matthewmcgarvey Could you provide failing code example?

matthewmcgarvey:
@Sija  https://sandbox.mint-lang.com/sandboxes/b7sVpfRO2_eJSQ

Namek:
The message could be improved. Can you suggest maybe, how would you see that error?

gdotdesign:
I would suggest something like:
 
```
You were trying to assign an incompatible value to the status state.

The type of the state is:

String

But the type you are trying to assign to it is:

Number

Here is where you did the assignment:

SNIPPET TO THE NEXT CALL

And here is where the state is defined:

SNIPPET TO THE STATE DEFINITION
```

PR #344, opened 2020-12-10
Better Error Message For `NextCallStateTypeMismatch`

s0kil:
Fixes https://github.com/mint-lang/mint/issues/339

### DIFF
--- a/src/messages/next_call_state_type_mismatch.cr
+++ b/src/messages/next_call_state_type_mismatch.cr
@@ -2,13 +2,15 @@ message NextCallStateTypeMismatch do
   title "Type Error"
 
   block do
-    text "The type of the state"
-    bold name
-    text "does not match of its definition in a"
-    bold "next call."
+    text "You were trying to assign an incompatible value to the status state."
+
+    text "\nThe type of the state is:"
+    bold expected
+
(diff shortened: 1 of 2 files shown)


src/messages/next_call_state_type_mismatch.cr
```diff
@@ -2,13 +2,15 @@ message NextCallStateTypeMismatch do
...
 
   block do
-    text "The type of the state"
-    bold name
-    text "does not match of its definition in a"
-    bold "next call."
-  end
+    text "You were trying to assign an incompatible value to the status state."
+
+    text "\nThe type of the state is:"
```
gdotdesign, line 7: Use a `block` block instead of a new line since these messages are shown in the browser as well and it will not honour the new line since it's not displayed in a `pre`.

gdotdesign (changes requested):

src/messages/next_call_state_type_mismatch.cr
s0kil: Got it.

gdotdesign (approved):

gdotdesign:
Great, thanks :+1:

PR #349, opened 2020-12-22
Add prefix to CSS classes

alex-lairan:
# Description

This fix the possibility to have a conflict between external CSS and Mint CSS.
Most of external css are compiled and use letter as identifier (as mint does).

~Skipping the 26 first reduce the chance that an important style apply.~

It allows the user to specify a specific prefix for css classes.
```json
  "application": {
    "css-prefix": "my_prefix-"
  },
```

### DIFF
--- a/core/tests/mint.json
+++ b/core/tests/mint.json
@@ -1,5 +1,8 @@
 {
   "name": "core-tests",
+  "application": {
+    "css-prefix": "titan-"
+  },
   "test-directories": [
     "tests"
   ]
--- a/core/tests/mint.json
+++ b/core/tests/mint.json
@@ -1,8 +1,5 @@
 {
   "name": "core-tests",
-  "application": {
-    "css-prefix": "titan-"
-  },
   "test-directories": [
     "tests"
   ]
--- a/spec/compilers/css_definition
+++ b/spec/compilers/css_definition
@@ -13,7 +13,7 @@ component Main {
 }
 --------------------------------------------------------------------------------
 class A extends _C {
-  $a() {
+  $aa() {
     const _ = {
       [`--a-a`]: this.a + `px 0px`
     };
@@ -27,8 +27,8 @@ class A extends _C {
 
   render() {
     return _h("div", {
-      className: `a`,
-      style: _style([this.$a()])
+      className: `aa`,
+      style: _style([this.$aa()])
     });
   }
 };
@@ -36,7 +36,7 @@ class A extends _C {
 A.displayName = "Main";
 
 _insertStyles(`
-.a {
+.aa {
   margin: var(--a-a);
 }
 `);
--- a/spec/compilers/css_definition
+++ b/spec/compilers/css_definition
@@ -13,7 +13,7 @@ component Main {
 }
 --------------------------------------------------------------------------------
 class A extends _C {
-  $foo-a() {
+  $a() {
     const _ = {
       [`--a-a`]: this.a + `px 0px`
     };
@@ -28,7 +28,7 @@ class A extends _C {
   render() {
     return _h("div", {
       className: `foo-a`,
-      style: _style([this.$foo-a()])
+      style: _style([this.$a()])
     });
   }
 };
--- a/spec/compilers/css_definition
+++ b/spec/compilers/css_definition
@@ -13,7 +13,7 @@ component Main {
 }
 --------------------------------------------------------------------------------
 class A extends _C {
-  $aa() {
+  $foo-a() {
     const _ = {
       [`--a-a`]: this.a + `px 0px`
     };
@@ -27,8 +27,8 @@ class A extends _C {
 
   render() {
     return _h("div", {
-      className: `aa`,
-      style: _style([this.$aa()])
+      className: `foo-a`,
+      style: _style([this.$foo-a()])
     });
   }
 };
@@ -36,7 +36,7 @@ class A extends _C {
 A.displayName = "Main";
 
 _insertStyles(`
-.aa {
+.foo-a {
   margin: var(--a-a);
 }
 `);
--- a/spec/compilers/css_definition
+++ b/spec/compilers/css_definition
@@ -27,7 +27,7 @@ class A extends _C {
 
   render() {
     return _h("div", {
-      className: `foo-a`,
+      className: `a`,
       style: _style([this.$a()])
     });
   }
@@ -36,7 +36,7 @@ class A extends _C {
 A.displayName = "Main";
 
 _insertStyles(`
-.foo-a {
+.a {
   margin: var(--a-a);
 }
 `);
--- a/src/builder.cr
+++ b/src/builder.cr
@@ -92,6 +92,9 @@ module Mint
     end
 
     def index
+      json = 
+        MintJson.parse_current
+
       runtime =
         Assets.read("runtime.js")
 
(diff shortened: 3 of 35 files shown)


Namek:
I totally agree for the alternative design. This one is just a workaround. I have seen some CSS framework which had classnames like `ee` so this PR wouldn't fix that case. Do you think you'd be able to implement the alternative?

alex-lairan:
I may try, do you have any hint for using variables from the `mint.json` file ? :)

Namek:
@gdotdesign can you advise?

alex-lairan:
I need to find a way to get values from `mint.json` and pass them into the `src/style_builder.cr` 

This class shouldn't read the json directly because it creates an unnecessary coupling.

gdotdesign:
I like the idea of having a prefix for this, however my personal choice would be command line flag, but the `application.css-prefix` can work too.

Code wise we should can pass the it from the top level command to the compiler itself.

alex-lairan:
I don't know why, but on my personal app, I don't have any html, but from manual test, and from specs, it works.
¯\_(ツ)_/¯

Can someone help me to find out why?

core/tests/mint.json
```diff
@@ -1,5 +1,8 @@
 {
   "name": "core-tests",
+  "application": {
+    "css-prefix": "titan-"
```
alex-lairan, line 4: I will change this, it's just for local testing purposes :)

spec/compilers/css_definition
```diff
@@ -13,7 +13,7 @@ component Main {
 }
 --------------------------------------------------------------------------------
 class A extends _C {
-  $a() {
+  $aa() {
```
alex-lairan, line 16: All of the `aa` will be removed and changed with a prefix.

I keep them to simplify a grep :)

alex-lairan:
I'm confused because `"css-prefix": "prefix-"` cause a silent error that lead to an empty web page, but `"css-prefix": "prefix_" works perfectly well.

gdotdesign:
> I'm confused because `"css-prefix": "prefix-"` cause a silent error that lead to an empty web page, but `"css-prefix": "prefix_" works perfectly well.

It's because the prefix contains a dash `-`. The selectors names are used to create the CSS variable map in the component which results in output like this:

```
$prefix-h() {
  const _ = {
    [`--a-a`]: `2px solid ` + this.aq,
    [`--a-b`]: this.ar
  };

  return _;
}
```

and this causes a JavaScript error. 

What we need to do is to just use the un-prefixed version internally (in JavaScript) and use the prefixed one for the actual classes.

src/builder.cr
```diff
@@ -119,7 +123,8 @@ module Mint
       end
 
       terminal.measure "  #{ARROW} Compiling: " do
-        compiled = Compiler.compile type_checker.artifacts, {optimize: true}
+        options = {optimize: true, css_prefix: @css_prefix}
+        compiled = Compiler.compile type_checker.artifacts, options
```
Sija, line 127:
```suggestion
        compiled = Compiler.compile type_checker.artifacts, {
          optimize:   true,
          css_prefix: @css_prefix,
        }
```

src/commands/compile.cr
```diff
@@ -45,7 +48,8 @@ module Mint
         end
 
         terminal.measure "  #{ARROW} Compiling: " do
-          compiled = Compiler.compile_embed type_checker.artifacts, {optimize: true}
+          options = {optimize: true, css_prefix: json.application.css_prefix}
+          compiled = Compiler.compile_embed type_checker.artifacts, options
```
Sija, line 52:
```suggestion
          compiled = Compiler.compile_embed type_checker.artifacts, {
            optimize:   true,
            css_prefix: json.application.css_prefix,
          }
```

src/builder.cr
```diff
@@ -1,8 +1,12 @@
 module Mint
   class Builder
+    @css_prefix : String
+
     def initialize(relative, skip_service_worker, skip_icons)
       json = MintJson.parse_current
 
+      @css_prefix = json.application.css_prefix || ""
```
Sija, line 8: `nil` is enough.

```suggestion
      @css_prefix = json.application.css_prefix
```

src/compiler.cr
```diff
@@ -11,9 +11,9 @@ module Mint
     @static_components = {} of String => String
     @static_components_pool = NamePool(String, Nil).new
 
-    def initialize(@artifacts : TypeChecker::Artifacts, @optimize = false)
+    def initialize(@artifacts : TypeChecker::Artifacts, @optimize = false, css_prefix = "")
```
Sija, line 14:
```suggestion
    def initialize(@artifacts : TypeChecker::Artifacts, @optimize = false, css_prefix = nil)
```

src/compilers/top_level.cr
```diff
@@ -2,15 +2,15 @@
 # in the compiler.
 module Mint
   class Compiler
-    alias Options = NamedTuple(optimize: Bool)
+    alias Options = NamedTuple(optimize: Bool, css_prefix: String)
```
Sija, line 5:
```suggestion
    alias Options = NamedTuple(optimize: Bool, css_prefix: String?)
```

src/compilers/top_level.cr
```diff
@@ -2,15 +2,15 @@
...
 
-    DEFAULT_OPTIONS = {optimize: false}
+    DEFAULT_OPTIONS = {optimize: false, css_prefix: ""}
```
Sija, line 7:
```suggestion
    DEFAULT_OPTIONS = {optimize: false, css_prefix: nil}
```

src/mint_json.cr
```diff
@@ -604,6 +609,19 @@ module Mint
       }
     end
 
+    # Parsing the css prefix
+    # --------------------------------------------------------------------------
+
+    json_error MintJsonCssPrefixInvalid
+
+    def parse_application_css_prefix
+      @parser.read_string
```
Sija, line 618:
```suggestion
      @parser.read_string_or_null
```

src/reactor.cr
```diff
@@ -65,7 +69,7 @@ module Mint
       type_checker.check
 
       # Compile.
-      @script = Compiler.compile type_checker.artifacts
+      @script = Compiler.compile type_checker.artifacts, {optimize: false, css_prefix: json.application.css_prefix}
```
Sija, line 72:
```suggestion
      @script = Compiler.compile type_checker.artifacts, {
        optimize:   false,
        css_prefix: json.application.css_prefix,
      }
```

src/style_builder.cr
```diff
@@ -124,7 +124,7 @@ module Mint
     getter selectors, property_pool, name_pool, style_pool, variables, ifs
     getter cases
 
-    def initialize
+    def initialize(@css_prefix = "")
```
Sija, line 127:
```suggestion
    def initialize(@css_prefix = nil)
```

src/style_builder.cr
```diff
@@ -189,10 +189,14 @@ module Mint
       false
     end
 
+    def prefixed_class_name(node)
+      @css_prefix + style_pool.of(node, nil)
```
Sija, line 193:
```suggestion
      @css_prefix.to_s + style_pool.of(node, nil)
```

Sija (changes requested):

src/compiler.cr
gdotdesign: why is `nil` better than an empty string `""`?

Sija: 1. `instance_sizeof(String) # => 16` vs `sizeof(Nil) # => 0`, not too much but still ;)
2. semantically speaking, `nil` IMO better signifies the _lack of_ value then an empty string

alex-lairan:
:tada: Thanks for the fix !

PR #362, opened 2021-01-23
Tweaks after Crystal 0.35 release

Sija:


src/installer.cr
```diff
@@ -141,8 +141,10 @@ module Mint
         unless @resolved[dependency]?
           eliminated =
             @eliminated
-              .select { |item| item[0][:name] == dependency }
-              .map { |item| "#{item[0][:version]} by #{item[2]} from #{item[1][:name]}:#{item[1][:version]}" }
+              .select(&.[0][:name].== dependency)
```
gdotdesign, line 144: I'm not sure but I'll always use parentheses in cases like this.

```suggestion
              .select(&.[0][:name].==(dependency))
```

gdotdesign (approved):

src/type_checkers/scope.cr
```diff
@@ -89,8 +89,8 @@ module Mint
 
       def current_top_level_entity?
         @levels.find do |item|
-          item.is_a?(Ast::Store) || 0
```
gdotdesign, line 92: This probably caused some bugs :sweat_smile: Nice catch!

PR #364, opened 2021-01-29
Support constant declarations in test suites

Sam647254:
This pull request enables constants to be declared directly in test suites; the syntax for declaring them is the same as in modules (please see the spec in this pull request for an example). Constants declared in a suite may be used in any test in that suite, but are not accessible in other suites in the same file.

Implementation details: The constants go into the `constants` object in the output `SUITE` JavaScript object (alongside the existing `name` and `tests`), and the test runner passes the constants as an argument to the `proc` function of each test (which used to accept no arguments). In the variable resolution step, the compiler checks whether the variable is an `Ast::Constant` whose parent is an `Ast::Suite`. If this is the case, the emitted variable name becomes `constants.#{name}()` instead of the usual `this.#{name}` (please see src/compilers/variable.cr for details).

Please let me know if my pull request needs any changes or improvements; feedback would be much appreciated!


gdotdesign (approved):

gdotdesign:
Very nice PR! :tada: Thank you very much!

I'll wait for @Sija before merging but it LGTM!

Sam647254:
Thank you for the review!

Sija (approved):
Looks great, just :shipit: :)

PR #365, opened 2021-01-29
Update To Latest Crystal 0.36.0

s0kil:
Also fixing this issue https://github.com/crystal-ameba/ameba/pull/184


shard.yml
```diff
@@ -42,6 +42,6 @@ targets:
   mint:
     main: src/mint.cr
 
-crystal: ">= 0.35.0"
+crystal: ">= 0.36.0"
```
Sija, line 45: That wasn't necessary btw.

PR #366, opened 2021-02-01
Gracefully shut down the test runner in case of uncaught exceptions in JavaScript

Sam647254:
Previously, when running tests, if the emitted JavaScript is invalid and causes an exception to be thrown (such as in the case of [this commit](https://github.com/mint-lang/mint/commit/98279f24858b0fb09a1893bb62aa4c2a68531de2), which fixed a bug in which missing parentheses caused a syntax error; thanks again for the fix!), the tests would stop running, but the test runner would still be running, still waiting for the `DONE` message that will never arrive, and so the test runner would be hanging forever. This pull request introduces a global handler that catches all unhandled exceptions in promises and sends the `CRASHED` message to the test runner, signalling an "internal error" and shutting down the test server and runner, instead of leaving the test runner hanging.

To test this, I simply renamed the `tests` field in a suite to `test` (while still keeping the `this.suite.tests` line to trigger a `TypeError` in JavaScript), and observed that running `mint test` will indeed display the expected exception message. I'm not sure how to set up a unit test for this handler, since ideally, it should never be used (but in case that the emitted JavaScript is somehow invalid, it provides a recovery mechanism).

Currently, only the error message in the JavaScript exception is displayed (as opposed to the entire stack trace). Please let me know if this (or any other aspects of the pull request) should be changed.

### DIFF
--- a/src/test_runner.cr
+++ b/src/test_runner.cr
@@ -49,17 +49,14 @@ module Mint
 
                 window.onerror = (message) => {
                   if (this.socket.readyState === 1) {
-                    console.log("!!1");
                     this.crash(message);
                   } else {
-                    console.log("!!2");
-                    error ||= message;
+                    error = error || message;
                   }
                 }
 
                 this.socket.onopen = () => {
                   if (error != null) {
-                    console.log("!!3");
                     this.crash(error);
                   }
 
@@ -80,7 +77,7 @@ module Mint
               run () {
                 return new Promise((resolve, reject) => {
                   this.next(resolve, reject)
-                }).then(() => this.socket.send("DONE"));
+                }).finally(() => this.socket.send("DONE"));
               }
 
               crash (message) {
@@ -363,7 +360,7 @@ module Mint
               terminal.puts "    |> #{failure.result}".colorize(:red)
             end
 
-           stop_server
+            stop_server
           else
             data = Message.from_json(message)
             case data.type
--- a/src/test_runner.cr
+++ b/src/test_runner.cr
@@ -82,21 +82,21 @@ module Mint
 
               next (resolve, reject) {
                 requestAnimationFrame(async () => {
-                  try {
-                    if (!this.suite || this.suite.tests.length === 0) {
-                      this.suite = this.suites.shift()
+                  if (!this.suite || this.suite.tests.length === 0) {
+                    this.suite = this.suites.shift()
 
-                      if (this.suite) {
-                        this.socket.send(JSON.stringify({ type: "SUITE", name: this.suite.name, result: "" }))
-                      } else {
-                        return resolve()
-                      }
+                    if (this.suite) {
+                      this.socket.send(JSON.stringify({ type: "SUITE", name: this.suite.name, result: "" }))
+                    } else {
+                      return resolve()
                     }
+                  }
 
-                    let test = this.suite.tests.shift()
+                  let test = this.suite.tests.shift()
 
-                    let currentHistory = window.history.length
+                  let currentHistory = window.history.length
 
+                  try {
                     let result = await test.proc(this.suite.constants)
 
                     // Go back to the beginning
@@ -124,12 +124,11 @@ module Mint
                         this.socket.send(JSON.stringify({ type: "FAILED", name: test.name, result: "false" }))
                       }
                     }
-
-                    this.next(resolve, reject)
                   } catch (error) {
                     // An error occurred while trying to run a test; this is different from the test itself failing.
-                    reject(error.toString());
+                    this.socket.send(JSON.stringify({ type: "ERRORED", name: test.name, result: error.toString() }));
                   }
+                  this.next(resolve, reject);
                 })
               }
             }
@@ -374,6 +373,9 @@ module Mint
               @succeeded += 1
             when "FAILED"
               @reporter.failed data.name, data.result
+              @failed << data
+            when "ERRORED"
+              terminal.puts "An error occurred when running the test #{data.name}: #{data.result}".colorize(:red)
               @failed << data
             when "CRASHED"
               @reporter.crashed data.result
--- a/src/test_runner.cr
+++ b/src/test_runner.cr
@@ -35,11 +35,7 @@ module Mint
                       result = value.toString()
                     }
 
-                    this.socket.send(JSON.stringify({
-                      result: result,
-                      name: "",
-                      type: "LOG",
-                    }))
+                    this.log(result);
                   }
                 }
 
@@ -57,10 +53,6 @@ module Mint
                   if (error != null) {
                     this.crash(error);
                   }
-
-                  window.addEventListener('unhandledrejection', (event) => {
-                    this.crash(event.reason);
-                  });
                 }
               }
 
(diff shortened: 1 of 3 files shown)


Sam647254:
Edit: I've rearranged the `script` blocks in the HTML, so that the `TestRunner` class is declared and instantiated before the tests are loaded, so that if the tests contain syntax errors, they will also be caught and reported to the test runner. When I first opened the pull request, the changes I made could only catch semantic errors (such as misnamed variables), but not syntax errors.

Sija (commented):
Nice! I left couple of comments. Also needs a `crystal tool format`

src/test_runner.cr
```diff
@@ -45,16 +45,46 @@ module Mint
...
 
+                let error = null;
+
+                window.onerror = (message) => {
+                  if (this.socket.readyState === 1) {
+                    console.log("!!1");
+                    this.crash(message);
+                  } else {
+                    console.log("!!2");
+                    error ||= message;
```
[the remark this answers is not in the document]
gdotdesign: Seems so: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR_assignment but it's kind of new.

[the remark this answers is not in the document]
Sija: I'd avoid it for compatibility reasons.

src/test_runner.cr
```diff
@@ -45,16 +45,46 @@ module Mint
...
+                  this.socket.addEventListener("open", this.run);
                 }
               }
 
-              async run () {
+              run () {
                 return new Promise((resolve, reject) => {
                   this.next(resolve, reject)
-                })
+                }).then(() => this.socket.send("DONE"));
```
Sija, line 83:
```suggestion
                }).finally(() => this.socket.send("DONE"));
```

src/test_runner.cr
```diff
@@ -45,16 +45,46 @@ module Mint
 
                 this.suites = suites
 
+                let error = null;
+
+                window.onerror = (message) => {
+                  if (this.socket.readyState === 1) {
+                    console.log("!!1");
```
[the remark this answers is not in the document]
Sam647254: Removed; I actually removed them before the force-push that you see above, but I made the silly mistake of forgetting to stage the changes (so the amend commit actually changed nothing); sorry about that!

src/test_runner.cr
```diff
@@ -45,16 +45,46 @@ module Mint
...
+                    this.crash(message);
+                  } else {
+                    console.log("!!2");
```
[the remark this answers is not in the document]
Sam647254: Removed

src/test_runner.cr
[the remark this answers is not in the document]
Sam647254: I've removed this usage.

src/test_runner.cr
```diff
@@ -45,16 +45,46 @@ module Mint
...
+                    error ||= message;
+                  }
+                }
+
                 this.socket.onopen = () => {
-                  this.run()
-                    .then(() => this.socket.send("DONE"))
+                  if (error != null) {
+                    console.log("!!3");
```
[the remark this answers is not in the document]
Sam647254: Removed

src/test_runner.cr
Sam647254: Ah, good idea!

Sam647254:
Thanks for the review and suggestions! Please let me know if anything else needs to be changed.

Sija:
@Sam647254 all of the core tests are hanging for some reason 🤔

src/test_runner.cr
```diff
@@ -43,18 +43,44 @@ module Mint
...
                 this.socket.onopen = () => {
-                  this.run()
-                    .then(() => this.socket.send("DONE"))
+                  if (error != null) {
+                    this.crash(error);
+                  }
+
+                  window.addEventListener('unhandledrejection', (event) => {
+                    event.promise.catch(e => this.crash(e.toString()));
+                  });
```
Sija, line 63: This I think should be moved out of `onopen` handler as it has nothing to do with it - could be put just right next to `window.onerror = ...` assignment.

Sam647254: Yes, that's possible, but since "crashing" means sending the `CRASHED` message through the socket, I purposefully placed it here since it's not possible for any unhandled rejections to occur until the tests have started (at which point the socket is already open and ready to receive messages). This way, it avoids the need to "cache" the error in a variable (like in the `window.onerror` handler, which catches errors that may occur before the socket is open, such as syntax errors).

Sam647254:
The stuck tests were caused by my (previously) problematic error handler, which was supposed to call `this.crash(e.message)` as opposed to just `crash(e.message)` (I forgot to add `this`; still in my Crystal mindset), thus causing the error handler to repeatedly call itself in an infinite recursion. It has been fixed now; sorry about that!

src/test_runner.cr
```diff
@@ -43,18 +43,44 @@ module Mint
...
 
                 this.socket.onopen = () => {
-                  this.run()
-                    .then(() => this.socket.send("DONE"))
+                  if (error != null) {
+                    this.crash(error);
+                  }
+
+                  window.addEventListener('unhandledrejection', (event) => {
+                    event.promise.catch(e => this.crash(e.toString()));
```
Sija, line 62: btw, shouldn't this be:

```suggestion
                    this.crash(event.reason);
```

src/test_runner.cr
Sija: I think we should avoid binding `unhandledrejection` - it may propagate errors which might be acceptable and shouldn't crash the runner - logging 'em might be useful though.

Sam647254: Any unhandled exception that can reach this handler will always interrupt the test execution; we can log the errors, but the test runner would still hang forever. Alternatively, if you know the specific exceptions that are harmless, we can create a passthrough list that will make the handler skip calling `this.crash` if the exception in question is in the list.

Sija: Aren't you confusing unhandled (promise) _rejection_ with unhandled _exception_?

src/test_runner.cr
Sija: I've noticed you've changed it back to `then`, why?

src/test_runner.cr
gdotdesign: We could need both since `onerror` doesn't catch unhandled promises. 
https://codepen.io/gdotdesign/pen/oNYXENp?editors=1010

I'm not sure that we should bind it to the runner or not, theoretically there are cases where it could fail to catch rejections for example in a stores initial state, but that wouldn't cause a  crash.

> I think we should avoid binding `unhandledrejection` - it may propagate errors which might be acceptable and shouldn't crash the runner - logging 'em might be useful though.

I don't think there are system promises that it can catch, but in any case we can log them.

Sam647254: If I understood correctly, [this page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch) shows that exceptions being thrown in a Promise context will result in a rejection, which, if unhandled by a `.catch` chained function, will propagate to this global handler (i.e. an unhandled exception in the Promise becomes an unhandled rejection when propagated outside).

Actually, I just realized that, since we now have a `finally` chained function, it will guarantee that the test runner shuts down even if an error occurs, which means that this handler no longer needs to crash the test runner anymore. I'll change it to just log the error instead. The `onerror` handler still needs to crash the test runner, though, since reaching it would mean that something serious (like a syntax error) has occurred.

Sam647254:
With my latest commit, I've removed the `unhandledrejection` handler with a `.catch` function chained after the Promise in the `run` method; however, I did have to wrap the code block inside `requestAnimationFrame` in a `try`/`catch` block, since the fact that the callback is `async` creates its own "Promise" inside the `run` Promise, whose `.catch` function cannot handle any exceptions being thrown in this inner block. The `catch` block will call `reject` (which was previously unused) to propagate the error to the `.catch` function. Please let me know if this is a reasonable solution!

src/test_runner.cr
```diff
@@ -76,58 +68,68 @@ module Mint
...
-                    this.suite = this.suites.shift()
+              log (message) {
+                this.socket.send(JSON.stringify({ type: "LOG", name: "", result: message }));
+              }
 
-                    if (this.suite) {
-                      this.socket.send(JSON.stringify({ type: "SUITE", name: this.suite.name, result: "" }))
-                    } else {
-                      return resolve()
+              next (resolve, reject) {
```
Sija, line 83: Shoudn't this be `async`?

```suggestion
              async next (resolve, reject) {
```

src/test_runner.cr
```diff
@@ -76,58 +68,68 @@ module Mint
...
+                        this.socket.send(JSON.stringify({ type: "FAILED", name: test.name, result: "false" }))
+                      }
                     }
-                  }
 
-                  this.next(resolve, reject)
+                    this.next(resolve, reject)
+                  } catch (error) {
+                    // An error occurred while trying to run a test; this is different from the test itself failing.
+                    reject(error.toString());
```
Sija, line 131: To signify erroring test I'd use sth like

```suggestion
                    this.socket.send(JSON.stringify({ type: "ERRORED", name: test.name, result: error.toString() }))
```

Sam647254: Done; to make the test name accessible, I moved the `try` block to surround the `let result = ...` line and everything after that instead. I've edited the `TestRunner` class to accept `ERRORED` messages as well, which will print the error message and count the test as a failure.

src/test_runner.cr
```diff
@@ -35,29 +35,52 @@ module Mint
...
+                  this.socket.addEventListener("open", () => this.run());
+                }
+              }
+
+              run () {
                 return new Promise((resolve, reject) => {
-                  this.next(resolve, reject)
-                })
+                  this.next(resolve, reject);
+                }).catch(e => this.log(e))
```
Sija, line 71:
```suggestion
                }).catch(e => this.log(e.reason))
```

gdotdesign:
Thanks! :rocket: :tada:

PR #360, opened 2021-01-18
Language Server

gdotdesign:
This PR implements the basics of a language server for more information check out the discussion :arrow_right: https://github.com/mint-lang/mint/discussions/270#discussioncomment-231306

--------------------------

Some implementation details:

- the formatter now doesn't store the AST
- there is now a `Workspace` class which is responsible for watching the changes of the files (mostly does what the reactor did)
- it is now possible to resolve a `mint.json` from any of the source files
- all of the nodes are saved (in parse order) on the root AST node
- the protocol itself is separatable into a different shard

PR #368, opened 2021-02-02
Followup to #360

Sija:
Followup to #360

### DIFF
--- a/src/artifact_cleaner.cr
+++ b/src/artifact_cleaner.cr
@@ -1,7 +1,7 @@
 module Mint
   class ArtifactCleaner
     def self.clean(clean_global : Bool = false)
-      artifacts = [".mint", "dist"]
+      artifacts = %w[.mint dist]
 
       if clean_global
         self.safe_delete(MINT_PACKAGES_DIR)
--- a/src/compilers/component.cr
+++ b/src/compilers/component.cr
@@ -88,8 +88,8 @@ module Mint
 
       constructor =
         unless constructor_body.empty?
-          js.function("constructor", ["props"]) do
-            constructor_body.unshift js.call("super", ["props"])
+          js.function("constructor", %w[props]) do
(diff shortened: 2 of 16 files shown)


src/ls/completion_item/component.cr
```diff
@@ -20,14 +20,14 @@ module Mint
                 workspace.type_checker.cache[property]?
 
               value =
-                case type.try(&.name)
-                when "String"
+                case type
+                when String
```
gdotdesign, line 24: This would break the functionality since the value of `cache` is not a Crystal type (https://github.com/mint-lang/mint/blob/master/src/type_checkers/types.cr), this was it always goes to the else branch.

src/ls/completion_item/component.cr
```diff
@@ -20,14 +20,14 @@ module Mint
...
                   if default == %("")
                     %("${#{index + 2}}")
                   else
                     %(${#{index + 2}:#{default}})
                   end
-                when "Array"
+                when Array
```
gdotdesign, line 30: Ditto.

gdotdesign (approved):
I see most of these could go into ambea right? maybe we should contribute there.

What I can see so far:
- `%w[]` for string only arrays
- `zero?` instead of `== 0`
- `[] of String` => `%w[]` (I not sure about that since `[] of String` is more readable)

gdotdesign (changes requested):
See comments in other review.

src/ls/completion_item/component.cr
Sija: I got you, thanks for the clarification! ftr, it's actually comparing `TypeChecker::Checkable#name` (whereas `Checkable` is `Type | Record | Variable`), not `type.class#name` as it would with my changes.

gdotdesign: Yes! Next time I'll be more specific :smile:

gdotdesign (approved):

PR #369, opened 2021-02-02
Fix spelling errors

Sija:
SSIA

### DIFF
--- a/core/source/Provider/Mutation.mint
+++ b/core/source/Provider/Mutation.mint
@@ -5,7 +5,7 @@ record Provider.Mutation.Subscription {
 }
 
 /*
-A provider to provid events when the DOM strucutre of the given
+A provider to provide events when the DOM structure of the given
 element changes.
 */
 provider Provider.Mutation : Provider.Mutation.Subscription {
--- a/src/messages/operation_pipe_ambiguous.cr
+++ b/src/messages/operation_pipe_ambiguous.cr
@@ -2,11 +2,11 @@ message OperationPipeAmbiguous do
   title "Type Error"
 
   block do
-    text "We cannot determine the order of the operands becuase the pipe makes it ambiguous."
+    text "We cannot determine the order of the operands because the pipe makes it ambiguous."
   end
 
   block do
-    text "You sould add parentheses around the right operands."
+    text "Wrap operands in parentheses to avoid ambiguity."
   end
 
   snippet node
--- a/spec/type_checking/open_module
+++ b/spec/type_checking/open_module
@@ -28,7 +28,7 @@ module Test {
 
 module Test {
   fun hello : String {
-    "Overriden...."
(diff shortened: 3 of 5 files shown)


src/messages/operation_pipe_ambiguous.cr
```diff
@@ -2,11 +2,11 @@ message OperationPipeAmbiguous do
   title "Type Error"
 
   block do
-    text "We cannot determine the order of the operands becuase the pipe makes it ambiguous."
+    text "We cannot determine the order of the operands because the pipe makes it ambiguous."
   end
 
   block do
-    text "You sould add parentheses around the right operands."
+    text "You should add parentheses around the right operand."
```
gdotdesign, line 9: This was intentional, it's not the right (as in side) but as the proper(?) intentional(?) or better how about `You should add parentheses around the right operation.`?

Sija: Ouch, I wouldn't guess xD By _right_ you mean RHS?

Sija: Maybe `You should add parentheses around the ambiguous operand.`?

gdotdesign: Not the RHS, let me give you an example: 
```
10 * 20 |> function() 
```

It can be both operations, I mean the one that is _correct_ in the scope of what it supposed to do. Don't know how to phrase it properly :sweat_smile:

Sija: I see... 😅 Maybe let's try avoiding the subject, whatcha think about sth like `Wrap operands in parentheses to avoid ambiguity`? ;)

gdotdesign: > I see... sweat_smile Maybe let's try avoiding the subject, whatcha think about sth like `Wrap operands in parentheses to avoid ambiguity`? ;)

Perfect! :tada:

core/source/Provider/Mutation.mint
```diff
@@ -5,7 +5,7 @@ record Provider.Mutation.Subscription {
 }
 
 /*
-A provider to provid events when the DOM strucutre of the given
+A provider to provid events when the DOM structure of the given
```
Sija, line 8: `provid` here is off as well but I'm unable to discover the original meaning of this xD

gdotdesign: It supposed to be `provide` (missed `e` on the end)

Sija: Uhmm, so it's really `A provider to provide [...]`? 😅 That was too tricky for me :P

gdotdesign: Yep :joy:

gdotdesign (approved):

Issue #380, opened 2021-02-25
`try` cannot return a Result directly


itsgreggreg:
This example from the guide does not compile: https://www.mint-lang.com/guide/reference/control-expressions/try

You can check it out here: https://sandbox.mint-lang.com/sandboxes/XkK_Ctlj-UR06g

It seems that mint is unwrapping the final `Result` and attempting to return the `User` directly. To make this code work you must nest `Results` in `try`s return like so:
```
Result.ok(
        Result.ok(
          {
            email = email,
            name = name
          }))
```

itsgreggreg:
I'm happy to work on this though I'm not sure which functionality is the intended. Would mint rather uwrap a final Result or return a final Result directly in a `try` block?

PR #381, opened 2021-02-26
`try` can return a result directly

itsgreggreg:
I took this test https://github.com/mint-lang/mint/blob/master/spec/compilers/try_with_catches_and_return_result and the docs as indication that it is the intention that `try` returns its last statement directly even if it is a Result.

This PR does 2 things:
1) `try` now directly returns its last statement even if it is a `Result`
2) If a non final statement in a `try` is a `Result` but its error type is a type variable then it is not catchable. I think this is fine, as I think this is only the case if the error is not possible but I'm not positive this is always the case in mint.

Happy to push this in what ever direction works for mint. By no means trying to influence the semantics of `try` just bug fixing. If the current functionality of `try` is the correct functionality and it is the docs that are wrong I'm happy to close this and fix the docs!

fixes: #380

### DIFF
--- a/spec/formatters/try_with_catch
+++ b/spec/formatters/try_with_catch
@@ -6,7 +6,7 @@ module Result {
 
 module A {
   fun test : String {
-    try{x=Result.error("hello")}catch String => a {"blah"}
+    try{x=Result.error("hello")x}catch String => a {"blah"}
   }
 }
 --------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


gdotdesign:
Thank you very much! :bow: 

It probably shouldn't unwrap results, it changed around a few times if I can remember (but probably by mistake).

The only thing is that it should be consistent with `parallel` and `sequence` as well (which unwraps `Result` as well as `Promise`). But that can be an other PR so I'll just merge this one.

PR #385, opened 2021-03-31
Rename manifest.json to manifest.webmanifest

jonathanoberg:
Specification suggest using .webmanifest in place of .json -- see https://www.w3.org/TR/appmanifest/ 

Also picked up by webhint lint -- see https://webhint.io/docs/user-guide/hints/hint-manifest-file-extension


gdotdesign:
Great! Thanks :+1:

PR #389, opened 2021-04-05
Implement asset directive.

gdotdesign:
This PR implements the `@asset` directive.

It allows referencing assets - images, stylesheets, videos, scripts, etc... in Mint code. It can be used as an expression and it inlines the URL to that asset.

Having this directive allows us to:

- Hash the assets based on their content so caching is easier
- Only include assets in the build that are touched by the type checker
- Deprecate `external.stylesheets` and `external.javascripts` in the `mint.json` 

-----

Example:
```
component Main {
  fun render : Html {
    <>
      <Html.Portals.Head>
        <link rel="stylesheet" href={@asset(some-css-file.css)}/>
        <script src={@asset(some-js-lib.js)} type="text/javascript"/>
      </Html.Portals.Head>
      <img src={@asset(some-image.jpg)}/>
    </>
  }
}
```


src/ast/directives/asset.cr
```diff
@@ -0,0 +1,41 @@
...
+                       @from : Int32,
+                       @to : Int32)
+        end
+
+        def real_path
+          Path[Path[input.file].dirname, path].expand
+        end
+
+        def filename
+          if exists?
```
Sija, line 18: Prefer early returns:

```suggestion
          return unless exists?
```

src/builder.cr
```diff
@@ -34,6 +37,18 @@ module Mint
...
 
+      unless artifacts.assets.empty?
+        FileUtils.mkdir Path[DIST_DIR, ASSET_DIR].to_s
+
+        terminal.measure "#{COG} Writing assets..." do
+          artifacts.assets.each do |asset|
+            File.write Path[DIST_DIR, ASSET_DIR, asset.filename.to_s], asset.real_path
+          end
+
+          File.write Path[DIST_DIR, "external-stylesheets.css"], SourceFiles.external_stylesheets
```
Sija, line 48: Shouldn't this be left outside of the `unless` branch?

src/messages/asset_directive_expected_file.cr
```diff
@@ -0,0 +1,13 @@
+message AssetDirectiveExpectedFile do
+  title "Syntax Error"
+
+  block do
+    text "The path specified for an asset directive does not exists."
+  end
+
+  block do
+    text "The file should be here: #{path}"
```
Sija, line 9:
```suggestion
    text "The file should be here:"
    bold path
```

src/reactor.cr
```diff
@@ -9,13 +9,15 @@ module Mint
...
     @error : String?
-    @ast : Ast = Ast.new
-    @script = ""
     @host : String
     @port : Int32
-    @auto_format : Bool
+
+    @sockets : Array(HTTP::WebSocket) = [] of HTTP::WebSocket
+    @script : String = ""
+    @ast : Ast = Ast.new
```
Sija, line 20: Inference to the rescue!

```suggestion
    @sockets = [] of HTTP::WebSocket
```

The removed declarations are repeated with the `getter` calls below.

src/reactor.cr
```diff
@@ -145,6 +149,22 @@ module Mint
         SourceFiles.external_stylesheets.to_s
       end
 
+      get "/assets/:name" do |env|
```
Sija, line 152: Shouldn't this use rather the asset path?

```suggestion
      get "/assets/*path" do |env|
```

src/builder.cr
gdotdesign: Yes, I was planning to fix this later :joy:

src/reactor.cr
gdotdesign: No the assets cannot be in nested paths so `:name` is sufficient, but it should use the `ASSET_DIR` constant.

src/compilers/directives/asset.cr
```diff
@@ -0,0 +1,7 @@
+module Mint
+  class Compiler
+    def _compile(node : Ast::Directives::Asset) : String
+      "`/assets/#{node.filename}`"
```
Sija, line 4:
```suggestion
      "`/#{ASSET_DIR}/#{node.filename}`"
```

src/ast/directives/asset.cr
```diff
@@ -0,0 +1,41 @@
...
+          return unless exists?
+
+          hash =
+            Digest::MD5.new
+              .update(File.read(real_path))
+              .final
+              .hexstring
+
+          extname =
+            File.extname(real_path)
```
Sija, line 27:
```suggestion
            real_path.extension
```

src/reactor.cr
```diff
@@ -145,6 +147,22 @@ module Mint
         SourceFiles.external_stylesheets.to_s
       end
 
+      get "/#{ASSET_DIR}/:name" do |env|
+        asset =
+          @artifacts.try(&.assets.find(&.filename.==(env.params.url["name"])))
```
Sija, line 152:
```suggestion
        filename =
          env.params.url["name"]

        asset =
          @artifacts.try(&.assets.find(&.filename.==(filename)))
```

src/reactor.cr
```diff
@@ -145,6 +147,22 @@ module Mint
...
+
+        next unless asset
+
+        # Set cache to expire in 30 days.
+        env.response.headers["Cache-Control"] = "max-age=2592000"
+
+        # Try to figure out mime type from name.
+        env.response.content_type =
+          MIME.from_filename?(env.params.url["name"]).to_s
```
Sija, line 161:
```suggestion
          MIME.from_filename?(filename).to_s
```

src/builder.cr
```diff
@@ -34,6 +38,17 @@ module Mint
...
       end
 
+      unless artifacts.assets.empty?
+        FileUtils.mkdir Path[DIST_DIR, ASSET_DIR].to_s
+
+        terminal.puts "#{COG} Writing assets..."
+
+        artifacts.assets.each do |asset|
+          puts "  #{ARROW} #{asset.filename}"
+          File.write Path[DIST_DIR, ASSET_DIR, asset.filename.to_s], asset.real_path
```
Sija, line 48: Shouldn't you write the original file contents, instead of its path?

```suggestion
          File.open(asset.real_path) do |io|
            File.write Path[DIST_DIR, ASSET_DIR, asset.filename.to_s], io
          end
```

Sija: Or, if we don't care about copying some borked permissions, this could be simplified to:

```suggestion
          File.copy asset.real_path, Path[DIST_DIR, ASSET_DIR, asset.filename.to_s]
```

gdotdesign: Yeah, I totally missed that :man_facepalming: Thanks! :+1: :1st_place_medal:

Sija: ftw, the specs are not properly testing that atm.

Issue #405, opened 2021-04-24
Feature request: mint lint command


Daniel-Knights:
No idea how doable this is, but, a command like `mint lint` that returns all errors in a Mint project would be really useful. 

Reason being, I'd like to add linting/diagnostics to the VSCode extension, but, without a command like this, I'd need to parse every document separately and essentially reimplement Mint's error-handling from scratch. I could hack something together using `mint compile`, but, being able to run `mint lint` and get the results in a clean, readable format would make things 10x easier and would mean that the extension stays up-to-date with Mint's compiler.

gdotdesign:
It's not that hard to implement a `lint` command, we just need to figure out what should it do.

In brief this is how the errors work in Mint currently: 

1. We try to parse the `mint.json` file and if it has errors we show that and stop.
2. We go through each source and if we encounter an error we show that and stop.
3. We try to type check the whole application and if we encounter an error we show that and stop.

So basically we are only showing one error at a time. 

I guess what you would like to have (and please confirm) is:

- the error from the `mint.json` file
- parse errors from all files
- the first type error if there is one

Other considerations: 

- it should have optionally a JSON output
- it should return exit code 1 of there are any errors so it can be included in CI/CD
- it can run checks for the formatting as well on any files that doesn't have any parse errors
- it should be a separate class for reusability (for the language server)

By the way this will be essentially the diagnostics information for the language server as well.

----------------

This is a good first issue for someone who wants to contribute to the project, the `compile` command can be a good reference: https://github.com/mint-lang/mint/blob/master/src/commands/compile.cr

Daniel-Knights:
That'd be perfect @gdotdesign. I would volunteer to make a pull request, but I don't have any experience with Crystal. I might fork the repo and give it a go, but no promises 😅

PR #406, opened 2021-04-24
Implement lint command feature

Daniel-Knights:
As I said before, I have no experience with Crystal, so this could all be completely useless. If anything, it's a start for someone else to pick up and finish off.

This PR addresses #405.

# What

Following `compile.cr` as an example, and considering the points raised in #405, I've made an implementation of a new command for Mint: `mint lint`.

# How

- boolean flag `--json`, writes the output as an array to the STDOUT
- any error in `mint.json` **is** logged, but **isn't** captured within the JSON output
- parses all files and outputs any error it finds
- outputs only the first type-error it finds
- `exit 1` if there are any errors

## Notes

It would be really useful if all type-errors could be outputted instead of only the first one, again, no idea if this is doable.

Any parse errors cause a type-error (in my test project at least) which is misleading:

Screenshot 2021-04-24 at 20 37 22

Not the end of the world, but worth noting.

### DIFF
--- a/src/commands/lint.cr
+++ b/src/commands/lint.cr
@@ -26,16 +26,16 @@ module Mint
         errors = [] of String
         sources = [] of String
 
+        ast =
+          Ast.new
+            .merge(Core.ast)
+
         begin
           sources =
             Dir.glob(SourceFiles.all)
         rescue ex
           ex_handler errors, ex
         end
-
-        ast =
-          Ast.new
-            .merge(Core.ast)
 
         sources.reduce(ast) do |memo, file|
           parsed = Parser.parse(file)
--- a/src/commands/lint.cr
+++ b/src/commands/lint.cr
@@ -10,12 +10,6 @@ module Mint
         default: false,
         required: false
 
-      define_flag output : String,
-        description: "The output filename",
-        default: "lint.json",
-        required: false,
-        short: "o"
-
       def run
         execute "Linting" do
           lint
@@ -23,8 +17,8 @@ module Mint
       end
 
       def lint
-        errors = [] of String
         sources = [] of String
+        errors = [] of Exception
 
         ast =
           Ast.new
@@ -47,7 +41,7 @@ module Mint
           ex_handler errors, ex
         end
 
-        if errors.size == 0
+        if errors.empty?
           type_checker =
             TypeChecker.new(ast)
 
@@ -65,7 +59,7 @@ module Mint
         end
 
         if flags.json
-          File.write(flags.output, errors)
+          puts errors.compact_map(&.message.presence).to_json
         end
 
         if errors.size > 0
@@ -74,9 +68,7 @@ module Mint
       end
 
       def ex_handler(errors, ex)
-        if ex.message.is_a?(String)
-          errors << ex.message.as(String).to_json
-        end
+        errors << ex
 
         puts ex
       end
--- a/src/commands/lint.cr
+++ b/src/commands/lint.cr
@@ -28,7 +28,7 @@ module Mint
           sources =
             Dir.glob(SourceFiles.all)
         rescue ex
-          ex_handler errors, ex
+          errors << ex
         end
 
         sources.reduce(ast) do |memo, file|
@@ -38,7 +38,7 @@ module Mint
             memo.merge parsed
           end
         rescue ex
-          ex_handler errors, ex
+          errors << ex
         end
 
         if errors.empty?
@@ -51,7 +51,7 @@ module Mint
             begin
               type_checker.check
             rescue ex
-              ex_handler errors, ex
+              errors << ex
             else
               done = true
             end
@@ -60,15 +60,11 @@ module Mint
 
         if flags.json
           puts errors.compact_map(&.message.presence).to_json
+        else
+          puts errors
         end
 
         exit(errors.empty? ? 0 : 1)
-      end
-
-      def ex_handler(errors, ex)
-        errors << ex
-
-        puts ex
       end
     end
   end
--- a/src/commands/lint.cr
+++ b/src/commands/lint.cr
@@ -33,9 +33,6 @@ module Mint
           Ast.new
             .merge(Core.ast)
 
-        runtime =
-          Assets.read("runtime.js")
-
         errors = [] of String
 
         terminal.measure "  #{ARROW} Parsing #{sources.size} source files... " do
--- a/src/commands/lint.cr
+++ b/src/commands/lint.cr
@@ -33,7 +33,9 @@ module Mint
 
         sources.reduce(ast) do |memo, file|
           begin
-            parsed = Parser.parse(file)
+            parsed =
+              Parser.parse(file)
+
             memo.merge(parsed)
           rescue ex
             errors << ex
--- a/src/commands/lint.cr
+++ b/src/commands/lint.cr
@@ -62,9 +62,7 @@ module Mint
           puts errors.compact_map(&.message.presence).to_json
         end
 
-        if errors.size > 0
-          exit 1
-        end
+        exit(errors.empty? ? 0 : 1)
       end
 
       def ex_handler(errors, ex)
--- a/src/commands/lint.cr
+++ b/src/commands/lint.cr
@@ -3,7 +3,7 @@ module Mint
     class Lint < Admiral::Command
       include Command
 
-      define_help description: "Lints the project"
+      define_help description: "Lints the project for syntax and type errors."
 
       define_flag json : Bool,
         description: "Output errors to a JSON file",
@@ -23,37 +23,45 @@ module Mint
       end
 
       def lint
-        json =
-          MintJson.parse_current
+        errors = [] of String
+        sources = [] of String
 
-        sources =
-          Dir.glob(SourceFiles.all)
+        begin
+          sources =
+            Dir.glob(SourceFiles.all)
+        rescue ex
+          ex_handler errors, ex
+        end
 
         ast =
           Ast.new
             .merge(Core.ast)
 
-        errors = [] of String
+        sources.reduce(ast) do |memo, file|
+          parsed = Parser.parse(file)
 
-        terminal.measure "  #{ARROW} Parsing #{sources.size} source files... " do
-          sources.reduce(ast) do |memo, file|
-            parsed = Parser.parse(file)
-
-            if memo
-              memo.merge parsed
-            end
-          rescue ex
-            ex_handler errors, ex
+          if memo
+            memo.merge parsed
           end
-        end
-
-        type_checker =
-          TypeChecker.new(ast)
-
-        terminal.measure "  #{ARROW} Type checking: " do
-          type_checker.check
         rescue ex
           ex_handler errors, ex
+        end
+
+        if errors.size == 0
+          type_checker =
+            TypeChecker.new(ast)
+
+          done = false
+
+          while !done
+            begin
+              type_checker.check
+            rescue ex
+              ex_handler errors, ex
+            else
+              done = true
+            end
+          end
         end
 
         if flags.json
--- a/src/commands/lint.cr
+++ b/src/commands/lint.cr
@@ -1,0 +1,80 @@
+module Mint
+  class Cli < Admiral::Command
+    class Lint < Admiral::Command
+      include Command
+
+      define_help description: "Lints the project"
+
+      define_flag json : Bool,
+        description: "Output errors to a JSON file",
+        default: false,
+        required: false
+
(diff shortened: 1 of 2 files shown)


gdotdesign:
Thanks for opening this PR! :tada:

> It would be really useful if all type-errors could be outputted instead of only the first one, again, no idea if this is doable.
> 
> Any parse errors cause a type-error (in my test project at least) which is misleading:

Since parse errors leads to not including the entities in that file it might result in a type error if any entity in that file is used elsewhere. We can just not type-check if we have parsing errors. 

> any error in `mint.json` **is** logged, but **isn't** written to the JSON file (no idea if there's a way around this)

It's the same kind of error as the other so it can be catched (I'll suggest it in the review)

src/commands/lint.cr
```diff
@@ -0,0 +1,77 @@
+module Mint
+  class Cli < Admiral::Command
+    class Lint < Admiral::Command
+      include Command
+
+      define_help description: "Lints the project"
```
gdotdesign, line 6:
```suggestion
      define_help description: "Lints the project for syntax and type errors."
```

src/commands/lint.cr
```diff
@@ -0,0 +1,77 @@
...
+
+      def run
+        execute "Linting" do
+          lint
+        end
+      end
+
+      def lint
+        json =
+          MintJson.parse_current
```
gdotdesign, line 27: The errors from the `mint.json` can be catched like so:
```suggestion
          begin
            MintJson.parse_current
          rescue ex
            ex_handler errors, ex
          end
```

gdotdesign (commented):
We can skip `terminal.measure` calls since it doesn't make any sense to print it at the bottom.

There is a linting issue in the CI build, you can run the tests and linting with the `make test` command.

src/commands/lint.cr
Daniel-Knights: This seemed to be getting overridden by:

```crystal
sources =
  Dir.glob(SourceFiles.all)
```

So I removed `MintJson.parse_current` and wrapped a `begin...rescue` around this instead

Daniel-Knights:
I've added a `while` loop to allow for multiple type-errors to be caught, just let me know if you'd rather it didn't do this 👍

gdotdesign:
> I've added a `while` loop to allow for multiple type-errors to be caught, just let me know if you'd rather it didn't do this +1

It actually detects all type errors this way? :open_mouth:

Daniel-Knights:
> > I've added a `while` loop to allow for multiple type-errors to be caught, just let me know if you'd rather it didn't do this +1
> 
> It actually detects all type errors this way? 😮

It's still limited to one type-error per file 😕, but it does capture the first for each of them

Daniel-Knights:
Actually, I think it might be catching multiple per file. Perhaps it depends on the kind of type-error 🤔

It's catching type-errors from different functions in the same file, but only the first type-error from each function:

```bash
Mint - Linting
░ TYPE ERROR ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░

The 1st argument to a function is causing a mismatch. 

The function is expecting the 1st argument to be: 

  Html.Event(
    bubbles: Bool,
    ...
    event: Html.NativeEvent)

Instead it is: 

  String

The type of the function is: 

  Function(
    Html.Event(
      bubbles: Bool,
      ...
    Void)

You tried to call it here: 

┌──┬ source/components/Exchange.mint:28:32 ────────────────────────────────────┐
│24│   }                                                                       │
│25│                                                                           │
│26│   fun handleSubmit (e : Html.Event) {                                     │
│27│     sequence {                                                            │
│28│       Html.Event.preventDefault("e")                                      │
│29│                                                                           │
│30│       if (baseValues.amount == "0") {                                     │
│31│         setStatusMessage(6)                                               │
│32│       } else {                                                            │
└──┴───────────────────────────────────────────────────────────────────────────┘


░ TYPE ERROR ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░

You were trying to assign an incompatible value to the status state. 

The type of the state is: String 

But the type you are trying to assign to it: Number 

Here is where you did the assignment: 

┌──┬ source/components/Exchange.mint:23:12 ────────────────────────────────────┐
│19│     next { baseValues = values }                                          │
│20│   }                                                                       │
│21│                                                                           │
│22│   fun updateCompareLocale (values : InputValues) : Promise(Never, Void) { │
│23│     next { compareLocale = 6 }                                            │
│24│   }                                                                       │
│25│                                                                           │
│26│   fun handleSubmit (e : Html.Event) {                                     │
│27│     sequence {                                                            │
└──┴───────────────────────────────────────────────────────────────────────────┘

And here is where the state is defined: 

┌──┬ source/components/Exchange.mint:16:3 ─────────────────────────────────────┐
│12│     display: flex;                                                        │
│13│   }                                                                       │
│14│                                                                           │
│15│   state baseValues : InputValues = InputValues("0", "GBP")                │
│16│   state compareLocale : String = "USD"                                    │
│17│                                                                           │
│18│   fun updateBaseValues (values : InputValues) : Promise(Never, Void) {    │
│19│     next { baseValues = values }                                          │
│20│   }                                                                       │
└──┴───────────────────────────────────────────────────────────────────────────┘
```

Daniel-Knights:
I'm also wondering if it's worth formatting the JSON better. So, have each error as an object with separate properties, one for error type, one for line/character number and one for message

src/commands/lint.cr
```diff
@@ -0,0 +1,85 @@
...
+
+        if errors.size > 0
+          exit 1
+        end
+      end
+
+      def ex_handler(errors, ex)
+        if ex.message.is_a?(String)
+          errors << ex.message.as(String).to_json
+        end
```
Sija, line 79: You can transform exceptions at the final stage.

```suggestion
        errors << ex
```

src/commands/lint.cr
```diff
@@ -0,0 +1,85 @@
...
+            rescue ex
+              ex_handler errors, ex
+            else
+              done = true
+            end
+          end
+        end
+
+        if flags.json
+          File.write(flags.output, errors)
```
Sija, line 68:
```suggestion
          puts errors.compact_map(&.message.presence).to_json
```

Sija (changes requested):

src/commands/lint.cr
```diff
@@ -0,0 +1,85 @@
...
+      define_flag json : Bool,
+        description: "Output errors to a JSON file",
+        default: false,
+        required: false
+
+      define_flag output : String,
+        description: "The output filename",
+        default: "lint.json",
+        required: false,
+        short: "o"
```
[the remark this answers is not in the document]
Daniel-Knights: What should I change here: `File.write(flags.output, errors)`?

src/commands/lint.cr
```diff
@@ -0,0 +1,85 @@
...
+          parsed = Parser.parse(file)
+
+          if memo
+            memo.merge parsed
+          end
+        rescue ex
+          ex_handler errors, ex
+        end
+
+        if errors.size == 0
```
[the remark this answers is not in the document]
Daniel-Knights: `unless errors.empty?` doesn't seem to work properly, it works if I do `if errors.empty?`, is that right?

src/commands/lint.cr
```diff
@@ -0,0 +1,85 @@
...
+
+          while !done
+            begin
+              type_checker.check
+            rescue ex
+              ex_handler errors, ex
+            else
+              done = true
+            end
+          end
```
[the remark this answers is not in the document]
Daniel-Knights: This gives me an error: `expecting identifier 'end', not 'rescue'`

src/commands/lint.cr
```diff
@@ -0,0 +1,85 @@
...
+        end
+
+        if flags.json
+          File.write(flags.output, errors)
+        end
+
+        if errors.size > 0
+          exit 1
+        end
```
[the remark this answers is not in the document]
Daniel-Knights: This also gives me an error: `expecting token ')', not ':'`

src/commands/lint.cr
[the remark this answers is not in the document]
Sija: Of course, thanks for correction 🙏

src/commands/lint.cr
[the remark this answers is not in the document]
Sija: I've corrected the example.

src/commands/lint.cr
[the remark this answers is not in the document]
gdotdesign: Yep it seems that `while` is not a standard block: https://play.crystal-lang.org/#/r/azfk

src/commands/lint.cr
```diff
@@ -0,0 +1,75 @@
...
+              ex_handler errors, ex
+            else
+              done = true
+            end
+          end
+        end
+
+        if flags.json
+          puts errors.compact_map(&.message.presence).to_json
+        end
```
gdotdesign, line 63: This should have an else branch where it prints all the errors, otherwise STDOUT will have all the error because of `ex_handler` and then the JSON.

src/commands/lint.cr
```diff
@@ -0,0 +1,75 @@
...
+
+        exit(errors.empty? ? 0 : 1)
+      end
+
+      def ex_handler(errors, ex)
```
gdotdesign, line 68: See the previous comment, you can just remove this and use `errors << ex` instead.

gdotdesign (changes requested):
I just found one issue which is basically the exceptions are printed to STDOUT even if the `json` flag is set, if fixed then it's good to go for me :+1:

gdotdesign:
> I'm also wondering if it's worth formatting the JSON better. So, have each error as an object with separate properties, one for error type, one for line/character number and one for message. I'm not sure there's a simple way to do that, but, just a thought

Yeah, it would be great if the errors would contain the source path, line and column numbers but it's a bit tricky to do now, so let's add that as an enchantment for the future.

src/commands/lint.cr
Daniel-Knights: Like this?
```crystal
if flags.json
  puts errors.compact_map(&.message.presence).to_json
else
  puts errors
end
```

gdotdesign: Almost, `errors.each { |error| puts error }` otherwise it will print the brackets and commas since it's an array.

Daniel-Knights: Gotcha, okay. I've added this 👍

gdotdesign (approved):

src/commands/lint.cr
```diff
@@ -0,0 +1,71 @@
...
+
+          while !done
+            begin
+              type_checker.check
+            rescue ex
+              errors << ex
+            else
+              done = true
+            end
+          end
```
Sija, line 58:
```suggestion
          loop do
            type_checker.check
          rescue ex
            errors << ex
          else
            break
          end
```

Sija (changes requested):

src/commands/lint.cr
Daniel-Knights: If I do this I get an error on `make development`: `undefined method 'merge' for Array(Exception) (compile-time type is (Array(Exception) | Mint::Ast))`

src/commands/lint.cr
```diff
@@ -0,0 +1,67 @@
...
+        end
+
+        sources.reduce(ast) do |memo, file|
+          parsed = Parser.parse(file)
+
+          if memo
+            memo.merge parsed
+          end
+        rescue ex
+          errors << ex
```
Sija, line 41:
```suggestion
          begin
            parsed = Parser.parse(file)
            memo.merge(parsed)
          rescue ex
            errors << ex
          end
          memo
```

src/commands/lint.cr
Sija: That's because the return value from `reduce` assigned to the *memo*, and since `errors << ex` returns `errors`, the _memo_ gets `Array(Exception) | Mint::Ast)` type.

src/commands/lint.cr
```diff
@@ -0,0 +1,67 @@
...
+        begin
+          sources =
+            Dir.glob(SourceFiles.all)
+        rescue ex
+          errors << ex
+        end
+
+        sources.reduce(ast) do |memo, file|
+          begin
+            parsed = Parser.parse(file)
```
Sija, line 36:
```suggestion
            parsed =
              Parser.parse(file)

```

PR #417, opened 2021-05-16
Assorted refactors

Sija:


src/compilers/parallel.cr
```diff
@@ -1,6 +1,6 @@
 module Mint
   class Compiler
-    protected def prefix(statement : Ast::Statement, value : String)
+    protected def prefix(_node : Ast::Parallel, statement : Ast::Statement, value : String)
```
gdotdesign, line 3: Is this just for overloading?

src/compilers/try.cr
```diff
@@ -1,36 +1,42 @@
 module Mint
   class Compiler
+    protected def prefix(_node : Ast::Try, statement : Ast::Statement, value : String)
```
gdotdesign, line 3: Isn't this the same as for `sequence` and `parallel`?

gdotdesign (commented):
I like the changes overall :+1: I just have two comments about the `prefix` method in the compiler which seems to be the same for `try`, `sequence` and `parallel`.

src/compilers/parallel.cr
Sija: Yup, otherwise the methods would overwrite each other due to the same signatures.

src/compilers/try.cr
Sija: It is indeed - for `sequence` and `try`, although I wanted to maintain a clear separation of their definitions - where would you declare the shared method - in `src/compilers/try.cr` or `src/compilers/sequence.cr`? It could live in `src/compiler.cr` but that would be far away from its actual usage...

src/compilers/parallel.cr
Sija: Instead of overloading we could use different names, like `try_prefix`, `parallel_prefix` and so on - if that's better.

src/compilers/try.cr
gdotdesign: The only difference is between `sequence`/`try` and `parallel` is a `const` keyword, so it would make sense to me to have only one implementation with a flag. 

That said it can stay as is for now, I'm planning to merge all of them into a single feature in the future anyway.

gdotdesign (approved):

PR #419, opened 2021-05-17
Allow index argument in for expressions.

### LINKED ISSUE #418: Add index to for loop when using arrays
If you need the index to perform actions on an array, it seems like you have to use `Array.mapWithIndex`. This means that you need to create a computed property when you need to use the index in the render function.
It would be nice to be able to get the index in a normal for loop, then you could simply get rid of the computed property completely.

Example: https://sandbox.mint-lang.com/sandboxes/bmHCM6s4Kn8RgA

gdotdesign:
This PR implements the `index` as the last argument for `for` expressions:

it's the second argument for `Set(a)` and `Array(a)`:

```
for (item, index of ["Item1", "Item2"]) {
  index
}
```

and the third argument for `Map(key, value)`:

```
for (key, value, index of Map.empty()) {
  key
} 
```

Resolves #418

### DIFF
--- a/src/compilers/for_expression.cr
+++ b/src/compilers/for_expression.cr
@@ -1,17 +1,33 @@
 module Mint
   class Compiler
     def _compile(node : Ast::For) : String
+      subject_type =
+        cache[node.subject]
+
       body =
         compile node.body
 
       subject =
         compile node.subject
 
-      arguments =
-        if node.arguments.size == 1
-          js.variable_of(node.arguments[0])
+      arguments, index_arg =
+        if (subject_type.name == "Array" && node.arguments.size == 1) ||
+           (subject_type.name == "Set" && node.arguments.size == 1) ||
+           (subject_type.name == "Map" && node.arguments.size == 2)
+          if node.arguments.size == 1
+            {js.variable_of(node.arguments[0]), nil}
+          else
+            {js.array(node.arguments.map { |arg| js.variable_of(arg) }), nil}
+          end
         else
-          js.array(node.arguments.map { |arg| js.variable_of(arg) })
+          if node.arguments.size == 2
+            {js.variable_of(node.arguments[0]), node.arguments[1]}
+          else
+            {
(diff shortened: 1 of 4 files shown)


src/compilers/for_expression.cr
```diff
@@ -22,18 +38,24 @@ module Mint
           JS
         end
 
+      index =
+        if index_arg
+          "const #{js.variable_of(index_arg.not_nil!)} = _i"
```
Sija, line 43:
```suggestion
          "const #{js.variable_of(index_arg)} = _i"
```

src/type_checkers/for_expression.cr
```diff
@@ -25,18 +25,22 @@ module Mint
 
       raise ForMapArgumentsMismatch, {
         "node" => node,
-      } if is_map && node.arguments.size != 2
+      } if is_map && ![2, 3].includes?(node.arguments.size)
```
Sija, line 28:
```suggestion
      } if is_map && !node.arguments.size.in?(2, 3)
```

src/type_checkers/for_expression.cr
```diff
@@ -25,18 +25,22 @@ module Mint
...
 
       raise ForArrayOrSetArgumentsMismatch, {
         "node" => node,
-      } if is_array_or_set && node.arguments.size != 1
+      } if is_array_or_set && ![1, 2].includes?(node.arguments.size)
```
Sija, line 32:
```suggestion
      } if is_array_or_set && !node.arguments.size.in?(2, 3)
```

src/compilers/for_expression.cr
gdotdesign: I thought there is no overload for `variable_of(variable : Nil)`

Sija: `index_arg` can't be `Nil` at this point.

gdotdesign: Yep I figured :facepalm:

Sija (approved):
LGTM

PR #421, opened 2021-05-18
Replace time_format shard with TimeFormat module

matthewmcgarvey:
[time_format](https://github.com/vladfaust/time_format.cr) is archived and does not support Crystal 1.0.0. While the shard is very small, Mint does not even use it all. I created a `Mint::TimeFormat` module that keeps the same API used in the project but contains only the code that Mint uses.


gdotdesign (approved):

Sija (approved):
Nice 👍

PR #423, opened 2021-05-18
Replace tree_template with ECR

### LINKED ISSUE #422: Replace tree_template dependency with ECR
https://github.com/gdotdesign/tree_template is not updated to support Crystal 1.0.0. It's owned by @gdotdesign so it would be trivial to get it updated but I was wondering if it might be better to consider switching to ECR?

It's definitely a preference thing with choosing to build HTML with Crystal, but since this functionality isn't for users of mint and is only used in 2 places ([here](https://github.com/mint-lang/mint/blob/29df622d06feb1ad822cec5400dac9153203bcae/src/sandbox_server.cr#L96-L113) and [here](https://github.com/mint-lang/mint/blob/29df622d06feb1ad822cec5400dac9153203bcae/src/utils/index_html.cr#L25-L107)) maybe it would be better for two simple reasons:
- It's included in Crystal's standard library so it will always be up to date and removes a dependency
- it's more performant than an HTML dsl in crystal

matthewmcgarvey:
Fixes #422 

`TreeTemplate` usage in `src/sandbox_server.cr` seemed to be simple enough to warrant replacing with a heredoc instead of an ECR template.

### DIFF
--- a/src/utils/index_html.cr
+++ b/src/utils/index_html.cr
@@ -19,92 +19,10 @@ module Mint
       end
     end
 
-    def to_s(io : IO)
-      application = json.application
-
-      template = TreeTemplate.new(formatter: TreeTemplate::PrettyFormatter) do |t|
-        t.doctype :html5
-
-        t.html do
-          t.head do
-            t.meta(charset: application.meta["charset"]? || "utf-8")
-
-            t.title application.title.to_s
-
-            t.link(rel: "manifest", href: path_for("manifest.webmanifest"))
-
-            application.meta.each do |name, content|
-              next if name == "charset"
-              t.meta(name: name, content: content)
-            end
-
-            t.meta(name: "theme-color", content: application.theme)
-
-            # Insert the extra head content
-            t.unsafe application.head
-
-            unless @no_icons || application.icon.empty?
-              ICON_SIZES.each do |size|
-                t.link(
-                  rel: "icon",
-                  type: "image/png",
-                  href: path_for("icon-#{size}x#{size}.png"),
-                  sizes: "#{size}x#{size}"
-                )
-              end
-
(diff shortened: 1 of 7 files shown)


src/utils/index.html.ecr
```diff
@@ -0,0 +1,54 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="<%= application.meta["charset"]? || "utf-8" %>" />
+    <title>TEST: <%= application.title.to_s %></title>
```
Sija, line 5:
```suggestion
    <title>TEST: <%= application.title %></title>
```

Sija (commented):
I'd also rename the file `src/utils/index.html.ecr` to `src/utils/index_html.ecr` - to match the Crystal class it's binded to.

src/utils/index_html.cr
```diff
@@ -19,92 +19,10 @@ module Mint
...
-            end
-          end
-        end
-      end
-      io << template.render
+    private def application
+      json.application
     end
+
+    ECR.def_to_s "#{__DIR__}/index.html.ecr"
```
Sija, line 26:
```suggestion
    ECR.def_to_s "#{__DIR__}/index_html.ecr"
```

Sija (changes requested):

Sija (approved):

Sija:
That was the last remaining bit needed to ditch the `--ignore-crystal-version` flag, thanks @matthewmcgarvey! 🙏

gdotdesign (approved):

Sija:
I wonder why the macOS  build took so much time...

Sija:
Ok, seems it's more or less the same as before.

PR #424, opened 2021-05-18
Remove unused string_inflection shard

matthewmcgarvey:
[string_inflection](https://github.com/mosop/string_inflection) is another dependency that has not been updated to support Crystal 1.0.0.

Conveniently, Mint doesn't actually use it. From what I can find, the dependency hasn't been used since [the very beginning](https://github.com/mint-lang/mint/commit/f53e8f71254075b3efbc4701e6b567ed0dd81c26#diff-cbaa8062ee0fcfb5af8adfd98f69af4766b9ad5bb5e79158abe8da34f6fd7de0R8-R9).

### DIFF
--- a/shard.lock
+++ b/shard.lock
@@ -40,10 +40,6 @@ shards:
     git: https://github.com/luislavena/radix.git
     version: 0.4.1
 
-  string_inflection:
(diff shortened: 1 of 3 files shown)


Sija:
Nice catch! 👀

Issue #226, opened 2020-03-21
Destructure more complicated data structures


Namek:
Let's examine this ugly function:

```
enum FieldType {
  Island(Number, ConnectionSizes)
  Connection(Number)
}

/* in a module */
addConnection =
(
  idx1 : Number,
  idx2 : Number,
  genState : GenerationState
) : GenerationState {
  case (Map.get(idx1, genState.fieldsMap)) {
    Maybe::Just isl1 =>
      case (isl1) {
        FieldType::Island idx conns1 =>
          case (Map.get(idx2, genState.fieldsMap)) {
            Maybe::Just isl2 =>
              case (isl2) {
                FieldType::Island idx conns2 =>
                  try {
                    /* TODO update the genState */
                    genState
                  }

                => genState
              }

            => genState
          }

        => genState
      }

    => genState
  }
}
```

I would like to destructure tuple of 2 Maybes containing enums and directly access internal values of those enums.

```
case ({Map.get(idx1, genState.fieldsMap), Map.get(idx2, genState.fieldsMap)}) {
  {Maybe::Just(FieldType::Island(_, conns1)), Maybe::Just(FieldType::Island(_, conns2))} =>
    try {
      /* TODO update the genState */
      genState
    }

  => genState
}
```

Currently even a Tuple of built-in types is not destructurable.

Namek:
However, not just enums are important. Another useful case would be to destructure an array inside Maybe.

```
dirs = case (maybeDirections) {
  Maybe::Just([...dirs]) => dirs
  => []
}
```

teggotic:
I can take it

teggotic:
@gdotdesign Could you help me a bit?
@gdotdesign I have an issue when try getting a type of Ast::EnumDestructure on the compile stage. https://github.com/teggotic/mint/blob/19a247703cd2814a669fda0a5c5d73d18ee8abfb/src/compilers/case_branch.cr#L78 lookups[match] works on first function call, but params are not found there.
This is the code Im trying to compile
```
enum A {
  B(D)
  C
}
enum D {
  E(String)
  F
}

component Main {
  fun test(str : String) : Html {
    <div>
      <{str}>
    </div>
  }

  fun isLoggedIn (userState : A) : Html {
    case (userState) {
      A::B (D::E str) => test(str)
      A::C => <span/>
    }
  }

  fun render() : Html {
    isLoggedIn(A::B(D::E("Hello world")))
  }
}
```
[...]

PR #426, opened 2021-05-19
Update enum destructuring to look like creating enums

matthewmcgarvey:
Related to https://github.com/mint-lang/mint/issues/226

In order to support nested destructuring of Enum's, Mint will have to be able to distinguish nesting. Tuple's and Array's are fine since they are comma delimited, but Enums are not. This will be a breaking change, but updating the destructuring syntax to look like how you create them allows for someone to implement the next step.

## Before

```mint
enum UserState {
  LoggedIn(User, Config)
  Visitor
}

fun isLoggedIn (userState : UserState) : Bool {
  case (userState) {
    UserState::LoggedIn user config => true /* space delimited */
    UserState::Visitor => false
  }
}
```

## After

```mint
enum UserState {
  LoggedIn(User, Config)
  Visitor
}

fun isLoggedIn (userState : UserState) : Bool {
  case (userState) {
    UserState::LoggedIn(user, config) => true /* Wrapped in parens, comma delimited */
    UserState::Visitor => false
  }
}
```

### DIFF
--- a/src/parsers/enum_destructuring.cr
+++ b/src/parsers/enum_destructuring.cr
@@ -2,6 +2,7 @@ module Mint
   class Parser
     syntax_error EnumDestructuringExpectedDoubleColon
     syntax_error EnumDestructuringExpectedOption
+    syntax_error EnumDestructuringExpectedClosingParentheses
 
     def enum_destructuring
       start do |start_position|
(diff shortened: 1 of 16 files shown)


src/parsers/enum_destructuring.cr
```diff
@@ -11,7 +12,17 @@ module Mint
...
+
+        if char! '('
+          parameters.concat list(
+            terminator: ')',
+            separator: ','
+          ) { type_variable }.compact
+
+          whitespace
+          char ')', EnumDestructuringExpectedClosingParentheses
+        end
```
matthewmcgarvey, line 25: Copied completely from the way Mint parses Enum initialization https://github.com/mint-lang/mint/blob/2659e3c793cbc8821da8c54258a80d575333efee/src/parsers/enum.cr#L18-L30

gdotdesign:
Thank you for this PR! I agree that this makes it more understandable and will open the way for nested destructuring :+1:

gdotdesign (approved):

PR #427, opened 2021-05-20
Abstract away differences between standards and quirks modes

Sija:
Nature of the issue is described in more detail within this SO thread: https://stackoverflow.com/questions/7435843/window-top-document-body-scrolltop-not-working-in-chrome-or-firefox

Affects `Window.scroll{Width,Height,Left,Top}` methods.

### DIFF
--- a/core/source/Window.mint
+++ b/core/source/Window.mint
@@ -41,22 +41,66 @@ module Window {
 
   /* Returns the scrollable height of the window in pixels. */
   fun scrollHeight : Number {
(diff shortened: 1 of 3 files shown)


Sija:
Remaining failures are due to the issue described in this SO thread -> https://stackoverflow.com/questions/7435843/window-top-document-body-scrolltop-not-working-in-chrome-or-firefox

gdotdesign (approved):

PR #429, opened 2021-05-21
[CI] Run core specs on both, firefox and chrome browsers

Sija:


Sija:
@gdotdesign I see couple of ways how we can approach these 2 failing tests:

1. Fix the tests, so they'd succeed on `chrome`
2. Add an approach of running the tests conditionally - only/except on some browsers
3. Remove/Comment out the tests
4. Use `continue-on-error: true` and just ignore 'em

I'd lean towards option 1 if possible, and option 3 otherwise.

gdotdesign:
I think we can remove those tests for now since the methods work just not in that context, after that we can try to create something that works in Chrome using maybe [dispatchEvent](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent).

gdotdesign (approved):

PR #431, opened 2021-05-22
Remove compact calls after many and list calls

matthewmcgarvey:
many and list already stop if they encounter a falsey/nil value. There's no need to return an Array of nilable items. The means a **TON** of `.compact` calls can be removed.


gdotdesign:
Nice catch!

PR #434, opened 2021-05-22
Set `Content-Type` header to `application/json` in `Http.jsonBody`

### LINKED ISSUE #432: Should `Http.jsonBody` also set `Content-type` header
Hello,

I stumbled across this while using the `Http.jsonBody` method. Would it be a good idea to set `Content-type` as `application/json` when `Http.jsonBody` is called?

https://github.com/mint-lang/mint/blob/df3ac8b3ef81a2f3a5e46ccc09d8a1b3cd16066e/core/source/Http.mint#L155

tkshnwesper:
Resolves #432.

### DIFF
--- a/core/source/Http.mint
+++ b/core/source/Http.mint
@@ -167,8 +167,8 @@ module Http {
           request.headers
           |> Array.push(
             {
-              value = "application/json",
-              key = "Content-Type"
+              key = "Content-Type",
+              value = "application/json"
             })
         }
     }
@@ -236,8 +236,8 @@ module Http {
           })
         |> Array.push(
           {
-            value = value,
-            key = key
+            key = key,
+            value = value
           })
     }
   }
--- a/core/source/Http.mint
+++ b/core/source/Http.mint
@@ -1,3 +1,9 @@
+/* Represents a HTTP header */
+record Http.Header {
+  key : String,
+  value : String
+}
+
 /* Represents an HTTP request. */
 record Http.Request {
   headers : Array(Http.Header),
@@ -216,6 +222,17 @@ module Http {
           `new Record({ value: #{value}, key: #{key} })`,
           request.headers)
     }
+  }
+
+  /*
+  Checks the prescence of a header with the given key.
+
+    Http.empty()
+    |> Http.header("Content-Type", "application/json")
+    |> Http.hasHeader("Content-Type")
+  */
+  fun hasHeader (key : String, request : Http.Request) : Bool {
+    Array.any((header : Http.Header) : Bool { header.key == key }, request.headers)
   }
 
   /*
--- a/core/source/Http.mint
+++ b/core/source/Http.mint
@@ -232,16 +232,7 @@ module Http {
         request.headers
         |> Array.reject(
           (header : Http.Header) : Bool {
-            Regexp.createWithOptions(
-              key,
-              {
-                caseInsensitive = true,
-                multiline = false,
-                unicode = false,
-                global = false,
-                sticky = false
-              })
-            |> Regexp.match(header.key)
+            String.toLowerCase(header.key) == String.toLowerCase(key)
           })
         |> Array.push(
           {
@@ -262,16 +253,7 @@ module Http {
     request.headers
     |> Array.any(
       (header : Http.Header) : Bool {
-        Regexp.createWithOptions(
-          key,
-          {
-            caseInsensitive = true,
-            multiline = false,
-            unicode = false,
-            global = false,
-            sticky = false
-          })
-        |> Regexp.match(header.key)
+        String.toLowerCase(header.key) == String.toLowerCase(key)
       })
   }
 
--- a/core/source/Http.mint
+++ b/core/source/Http.mint
@@ -236,7 +236,21 @@ module Http {
     |> Http.hasHeader("Content-Type")
   */
   fun hasHeader (key : String, request : Http.Request) : Bool {
-    Array.any((header : Http.Header) : Bool { header.key == key }, request.headers)
+    Array.any(
+      (header : Http.Header) : Bool {
+        Regexp.match(
+          header.key,
+          Regexp.createWithOptions(
+            key,
+            {
+              caseInsensitive = true,
+              multiline = true,
+              unicode = true,
+              global = true,
+              sticky = true
+            }))
+      },
+      request.headers)
   }
 
   /*
--- a/core/source/Http.mint
+++ b/core/source/Http.mint
@@ -216,7 +216,7 @@ module Http {
   }
 
   /*
-  Adds a header to the request with the given key and value.
+  Adds a header to the request with the given key and value. Overwrites the value if key already exists.
 
     Http.empty()
     |> Http.header("Content-Type", "application/json")
@@ -224,9 +224,21 @@ module Http {
   fun header (key : String, value : String, request : Http.Request) : Http.Request {
     { request |
       headers =
-        Array.push(
-          `new Record({ value: #{value}, key: #{key} })`,
-          request.headers)
+        request.headers
+        |> Array.reject(
+          (header : Http.Header) : Bool {
+            Regexp.createWithOptions(
+              key,
+              {
+                caseInsensitive = true,
+                multiline = false,
+                unicode = false,
+                global = false,
+                sticky = false
+              })
+            |> Regexp.match(header.key)
+          })
+        |> Array.push(`new Record({ value: #{value}, key: #{key} })`)
     }
   }
 
--- a/core/source/Http.mint
+++ b/core/source/Http.mint
@@ -164,7 +164,12 @@ module Http {
         if (hasHeader("Content-Type", request)) {
           request.headers
         } else {
-          Array.push(`new Record({ value: "application/json", key: "Content-Type" })`, request.headers)
+          request.headers
+          |> Array.push(
+            {
+              value = "application/json",
+              key = "Content-Type"
+            })
         }
     }
   }
@@ -238,7 +243,11 @@ module Http {
               })
             |> Regexp.match(header.key)
           })
-        |> Array.push(`new Record({ value: #{value}, key: #{key} })`)
+        |> Array.push(
+          {
+            value = value,
+            key = key
+          })
     }
   }
 
--- a/core/source/Http.mint
+++ b/core/source/Http.mint
@@ -246,10 +246,10 @@ module Http {
             key,
             {
               caseInsensitive = true,
-              multiline = true,
-              unicode = true,
-              global = true,
-              sticky = true
+              multiline = false,
+              unicode = false,
+              global = false,
+              sticky = false
             }))
       },
       request.headers)
--- a/core/source/Http.mint
+++ b/core/source/Http.mint
@@ -160,7 +160,11 @@ module Http {
   fun jsonBody (body : Object, request : Http.Request) : Http.Request {
     { request |
       body = `JSON.stringify(#{body})`,
-      headers = Array.push(`new Record({ value: "application/json", key: "Content-Type" })`, request.headers) }
+      headers = if (hasHeader("Content-Type", request)) {
+        request.headers
+      } else {
+        Array.push(`new Record({ value: "application/json", key: "Content-Type" })`, request.headers)
+      } }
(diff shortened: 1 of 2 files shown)


core/source/Http.mint
```diff
@@ -168,7 +170,9 @@ module Http {
     |> Http.send()
   */
   fun formDataBody (body : FormData, request : Http.Request) : Http.Request {
-    { request | body = `#{body}` }
+    { request |
+      body = `#{body}`,
+      headers = [...request.headers, { key = "content-type", value = "multipart/form-data" }] }
```
Sija, line 175: This is specifically recommended **against** in the [MDN article](https://developer.mozilla.org/en-US/docs/Web/API/FormData/Using_FormData_Objects).

Sija (changes requested):

core/source/Http.mint
gdotdesign: Oh right so the browser sets it by default.

tkshnwesper: Cool, I removed it 👍

core/source/Http.mint
```diff
@@ -152,7 +152,9 @@ module Http {
     |> Http.send()
   */
   fun jsonBody (body : Object, request : Http.Request) : Http.Request {
-    { request | body = `JSON.stringify(#{body})` }
+    { request |
+      body = `JSON.stringify(#{body})`,
+      headers = Array.push(`new Record({ value: "application/json", key: "content-type" })`, request.headers) }
```
Sija, line 157:
```suggestion
      headers = Array.push(`new Record({ value: "application/json", key: "Content-Type" })`, request.headers) }
```

tkshnwesper: This crossed my mind too, but it appears to be [case-insensitive](https://stackoverflow.com/a/7718542/2513269). Or is this to improve the readability of the code?

Sija: I suggested it due to the conventional form in writing.

tkshnwesper: Point taken 👍

core/source/Http.mint
```diff
@@ -152,7 +152,9 @@ module Http {
     |> Http.send()
   */
   fun jsonBody (body : Object, request : Http.Request) : Http.Request {
-    { request | body = `JSON.stringify(#{body})` }
+    { request |
+      body = `JSON.stringify(#{body})`,
+      headers = Array.push(`new Record({ value: "application/json", key: "Content-Type" })`, request.headers) }
```
Sija, line 157: I think we should check for an already existing `Content-Type` record and skip the change if set or update it otherwise.

core/tests/tests/Http.mint
```diff
@@ -84,6 +84,19 @@ suite "Http.formDataBody" {
   }
 }
 
+suite "Http.jsonBody" {
```
Sija, line 87: I'd like to see 2 more test cases:
- testing that the existing headers remain intact
- testing that the `Content-Type` record remains intact if set before-hand

Sija (changes requested):

core/source/Http.mint
tkshnwesper: Good catch @Sija. 

Should we also handle the case for when the user adds the header after calling `jsonBody` by adding a check in the `header` function?

```mint
Http.empty()
|> Http.jsonBody(encode { user = "spaceman" }))
|> Http.header("Content-type", "application/json")
```

An alternate way of handling this could be to remove duplicates before making the actual HTTP call in the `send` function.

Sija: First, I'd fix the `Http.header` to take existing entries into account - i.e. overwriting them if found, and adding them otherwise. We could also add an additional argument to possibly skip overwriting, WDYT @gdotdesign?

gdotdesign: > First, I'd fix the `Http.header` to take existing entries into account - i.e. overwriting them if found, and adding them otherwise.

Overwriting existing headers sounds good :+1: 

> We could also add an additional argument to possibly skip overwriting, 

I would add a separate function instead, something like `Http.appendHeader(name, value)` which would just append and this way it wouldn't be a breaking change.

tkshnwesper: Yes, let's add the ability to overwrite the value if key already exists in `Http.header`.

On a different note, do you think the functionality to skip overwriting if the key already exists is really needed? I mean, if the user is setting the header then they should know if a key is being set or not and if they really mean to find out, they could always check the contents of the `Http.Request` record (that they themselves have created).

If we do decide to go with it, does `Http.appendHeader(name, value)` give the right idea about internal implementation? This might seem longer but gives a more clear idea:
- `Http.appendHeaderIfNotExists(name, value)` or 
- `Http.addHeaderIfNotExists(name, value)`

Just my 2 cents, what do you guys feel?

Sija: > I would add a separate function instead, something like `Http.appendHeader(name, value)` which would just append and this way it wouldn't be a breaking change.

Perhaps it would be a better API to add `fun hasHeader(key : String) : Bool` so the user can check whether given header *key* exists and decide by himself what's the more appropriate.

With such API design this function would need to just do (pseudo-code)

```mint
if (!Http.hasHeader("Content-Type", request)) {
  request = Http.header("Content-Type", "application/json", request)
}
request
```

tkshnwesper: That could also be a possibility, and it's a one-liner in mint code:

```
Array.any((header : Http.Header) : Bool { header.key == key }, request.headers)
```

tkshnwesper:
There's a little bit of code duplication in `Http.header` and `Http.hasHeader` functions. Shall we extract the common anonymous function into a different function? If yes, then where shall I put it?

core/tests/tests/Http.mint
tkshnwesper: > testing that the existing headers remain intact

Have added this to an existing test. See [`81cba8b` (#434)](https://github.com/mint-lang/mint/pull/434/commits/81cba8bcdbb825209b6f4bea59b5f0702977937f)

> testing that the Content-Type record remains intact if set before-hand

Created a new test for this. See [`6a7af43` (#434)](https://github.com/mint-lang/mint/pull/434/commits/6a7af43fa7c0539ceca90eb1bf6bebdfbb9c9298)

core/source/Http.mint
```diff
@@ -152,7 +158,20 @@ module Http {
...
+      headers =
+        if (hasHeader("Content-Type", request)) {
+          request.headers
+        } else {
+          request.headers
+          |> Array.push(
+            {
+              value = "application/json",
+              key = "Content-Type"
+            })
```
Sija, line 172:
```suggestion
          request |> Http.header("Content-Type, "application/json")
```

core/source/Http.mint
```diff
@@ -152,7 +158,20 @@ module Http {
     |> Http.send()
   */
   fun jsonBody (body : Object, request : Http.Request) : Http.Request {
-    { request | body = `JSON.stringify(#{body})` }
+    { request |
+      body = `JSON.stringify(#{body})`,
+      headers =
+        if (hasHeader("Content-Type", request)) {
+          request.headers
```
Sija, line 165:
```suggestion
          request
```

core/source/Http.mint
```diff
@@ -202,20 +221,60 @@ module Http {
...
+            Regexp.createWithOptions(
+              key,
+              {
+                caseInsensitive = true,
+                multiline = false,
+                unicode = false,
+                global = false,
+                sticky = false
+              })
+            |> Regexp.match(header.key)
```
Sija, line 244:
```suggestion
            String.toLowerCase(header.key) == String.toLowerCase(key)
```

core/source/Http.mint
```diff
@@ -202,20 +221,60 @@ module Http {
...
+          })
+        |> Array.push(
+          {
+            value = value,
+            key = key
```
Sija, line 249:
```suggestion
            key = key
            value = value
```

Sija (changes requested):

core/source/Http.mint
```diff
@@ -202,20 +221,60 @@ module Http {
...
+        Regexp.createWithOptions(
+          key,
+          {
+            caseInsensitive = true,
+            multiline = false,
+            unicode = false,
+            global = false,
+            sticky = false
+          })
+        |> Regexp.match(header.key)
```
Sija, line 274:
```suggestion
        String.toLowerCase(header.key) == String.toLowerCase(key)
```

core/tests/tests/Http.mint
```diff
@@ -84,6 +84,48 @@ suite "Http.formDataBody" {
...
+      method = "GET",
+      headers =
+        [
+          {
+            key = "existing header",
+            value = "value"
+          },
+          {
+            value = "application/json",
+            key = "Content-Type"
```
Sija, line 103:
```suggestion
            key = "Content-Type",
            value = "application/json"
```

core/tests/tests/Http.mint
```diff
@@ -84,6 +84,48 @@ suite "Http.formDataBody" {
...
+    |> Http.header("Content-Type", "text/plain")
+    |> Http.jsonBody(encode { user = "spaceman" })) == {
+      withCredentials = false,
+      body = `"{\"user\":\"spaceman\"}"`,
+      method = "GET",
+      headers =
+        [
+          {
+            value = "text/plain",
+            key = "Content-Type"
```
Sija, line 121:
```suggestion
            key = "Content-Type",
            value = "text/plain"
```

Sija (changes requested):

core/tests/tests/Http.mint
Sija: Great! 🙏

core/source/Http.mint
```diff
@@ -202,20 +221,60 @@ module Http {
...
     Http.empty()
     |> Http.header("Content-Type", "application/json")
   */
   fun header (key : String, value : String, request : Http.Request) : Http.Request {
     { request |
       headers =
-        Array.push(
-          `new Record({ value: #{value}, key: #{key} })`,
-          request.headers)
+        request.headers
```
Sija, line 232:
```suggestion
        request
```

core/source/Http.mint
gdotdesign: I think it's fine, we are setting the headers here.

core/source/Http.mint
gdotdesign: Ditto as before

core/source/Http.mint
Sija: Shouldn't the `request` be returned from here in order to allow chaining?

tkshnwesper: @Sija The value of `headers` is the result of an `if` expression. In a simpler form, it's basically doing the following:

```mint
fun jsonBody (body : Object, request : Http.Request) : Http.Request {
  { request |
    body = `JSON.stringify(#{body})`,
    headers = if (/* header exists */) { request.headers } else { /* insert into headers */ }
  }
}
```

core/source/Http.mint
tkshnwesper: Does this look alright?

```mint
fun jsonBody (body : Object, request : Http.Request) : Http.Request {
  if (hasHeader("Content-Type", request)) {
    { request | body = `JSON.stringify(#{body})` }
  } else {
    { request | body = `JSON.stringify(#{body})` }
    |> Http.header("Content-Type", "application/json")
  }
}
```

core/source/Http.mint
Sija: Now I see it, sorry for the confusion 🙏

Sija (approved):
LGTM, thanks @tkshnwesper! 🙏

core/source/Http.mint
Sija: Yup, looks fine, WDYT @gdotdesign ?

gdotdesign: LGTM :+1:

gdotdesign (approved):

core/source/Http.mint
tkshnwesper: Have made the change in #438

gdotdesign: Thanks!

PR #437, opened 2021-05-23
Can destructure nested tuples within tuples

matthewmcgarvey:
## Summary

This **ONLY** provides the ability to destructure a tuple inside a tuple.

```mint
component Main {
  fun render : String {
    case ({"A", {"B"}}) {
      {a, {b}} =>
        b
    }
  }
}
```

## Next Steps

The next changes/prs will be to expand the tuple destructuring to support nested enums and arrays. Then I will add the functionality to allow nested destructurings within those as well.

### DIFF
--- a/src/compilers/variable.cr
+++ b/src/compilers/variable.cr
@@ -96,12 +96,17 @@ module Mint
           else
             "SHOULD NEVER HAPPEN"
           end
-        when Tuple(Ast::Node, Int32)
+        when Tuple(Ast::Node, Int32), Tuple(Ast::Node, Array(Int32) | Int32)
           case item = entity[0]
           when Ast::WhereStatement, Ast::Statement
             case target = item.target
             when Ast::TupleDestructuring
-              js.variable_of(target.parameters[entity[1]])
+              case val = entity[1]
+              when Int32
+                js.variable_of(target.parameters[val])
+              when Array(Int32)
+                js.variable_of(val.reduce(target) { |curr_type, curr_val| curr_type.as(Ast::TupleDestructuring).parameters[curr_val] })
+              end.not_nil!
             else
               js.variable_of(node)
             end
--- a/src/compilers/variable.cr
+++ b/src/compilers/variable.cr
@@ -96,7 +96,7 @@ module Mint
           else
             "SHOULD NEVER HAPPEN"
           end
-        when Tuple(Ast::Node, Int32), Tuple(Ast::Node, Array(Int32) | Int32)
+        when Tuple(Ast::Node, Array(Int32) | Int32)
           case item = entity[0]
           when Ast::WhereStatement, Ast::Statement
             case target = item.target
--- a/src/compilers/variable.cr
+++ b/src/compilers/variable.cr
@@ -105,7 +105,9 @@ module Mint
               when Int32
                 js.variable_of(target.parameters[val])
               when Array(Int32)
-                js.variable_of(val.reduce(target) { |curr_type, curr_val| curr_type.as(Ast::TupleDestructuring).parameters[curr_val] })
+                js.variable_of(val.reduce(target) do |curr_type, curr_val|
+                  curr_type.as(Ast::TupleDestructuring).parameters[curr_val]
+                end)
               end.not_nil!
             else
               js.variable_of(node)
--- a/src/compilers/variable.cr
+++ b/src/compilers/variable.cr
@@ -102,13 +102,13 @@ module Mint
             case target = item.target
             when Ast::TupleDestructuring
               case val = entity[1]
-              when Int32
+              in Int32
                 js.variable_of(target.parameters[val])
-              when Array(Int32)
+              in Array(Int32)
                 js.variable_of(val.reduce(target) do |curr_type, curr_val|
                   curr_type.as(Ast::TupleDestructuring).parameters[curr_val]
                 end)
-              end.not_nil!
+              end
             else
               js.variable_of(node)
             end
--- a/src/errors/error.cr
+++ b/src/errors/error.cr
@@ -3,8 +3,8 @@ module Mint
   # to render to both to the Terminal and to an HTML format.
   class Error < Exception
     alias Value = String | Ast::Node | TypeChecker::Checkable |
-                  Array(TypeChecker::Checkable) | Tuple(Ast::Node, Int32) |
-                  Array(String) | Tuple(Ast::Node, Array(Int32))
+                  Array(TypeChecker::Checkable) | Tuple(Ast::Node, Int32 | Array(Int32)) |
+                  Array(String)
 
     alias Locals = Hash(String, Value)
 
--- a/src/errors/error.cr
+++ b/src/errors/error.cr
@@ -4,7 +4,7 @@ module Mint
   class Error < Exception
     alias Value = String | Ast::Node | TypeChecker::Checkable |
                   Array(TypeChecker::Checkable) | Tuple(Ast::Node, Int32) |
-                  Array(String)
+                  Array(String) | Tuple(Ast::Node, Array(Int32))
 
     alias Locals = Hash(String, Value)
 
--- a/src/errors/top_level.cr
+++ b/src/errors/top_level.cr
@@ -3,7 +3,7 @@ def raise(error : Mint::Error.class)
   raise error, {} of String => Mint::Error::Value
 end
 
-def raise(error : Mint::Error.class, raw : Hash(String, T)) forall T
+def raise(error : Mint::Error.class, raw : Hash(String, Mint::Error::Value?))
   locals = {} of String => Mint::Error::Value
 
   raw.map do |key, value|
--- a/src/errors/top_level.cr
+++ b/src/errors/top_level.cr
@@ -3,7 +3,7 @@ def raise(error : Mint::Error.class)
   raise error, {} of String => Mint::Error::Value
 end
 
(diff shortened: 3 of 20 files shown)


matthewmcgarvey:
Variable handling was by far the hardest part. It's very ugly but I really wanted to get it working on thought someone else might have a good idea on a better solution.

src/compilers/variable.cr
```diff
@@ -96,12 +96,17 @@ module Mint
...
           case item = entity[0]
           when Ast::WhereStatement, Ast::Statement
             case target = item.target
             when Ast::TupleDestructuring
-              js.variable_of(target.parameters[entity[1]])
+              case val = entity[1]
+              when Int32
+                js.variable_of(target.parameters[val])
+              when Array(Int32)
+                js.variable_of(val.reduce(target) { |curr_type, curr_val| curr_type.as(Ast::TupleDestructuring).parameters[curr_val] })
```
Sija, line 108:
```suggestion
                js.variable_of(val.reduce(target) do |curr_type, curr_val|
                  curr_type.as(Ast::TupleDestructuring).parameters[curr_val]
                end)
```

src/errors/top_level.cr
```diff
@@ -3,7 +3,7 @@ def raise(error : Mint::Error.class)
   raise error, {} of String => Mint::Error::Value
 end
 
-def raise(error : Mint::Error.class, raw : Hash(String, T)) forall T
+def raise(error : Mint::Error.class, raw : Hash(String, Mint::Error::Value?))
```
Sija, line 6: Why is this needed?

Sija (commented):
I have just a couple of comments, but besides that looks good!
Is this PR finished btw?

src/errors/top_level.cr
matthewmcgarvey: It was just a compilation error change I was making along the way. Forgot to revert it.

src/compilers/variable.cr
```diff
@@ -96,12 +96,19 @@ module Mint
           else
             "SHOULD NEVER HAPPEN"
           end
-        when Tuple(Ast::Node, Int32)
+        when Tuple(Ast::Node, Int32), Tuple(Ast::Node, Array(Int32) | Int32)
```
gdotdesign, line 99: `Tuple(Ast::Node, Array(Int32) | Int32)` is enough here since it's matching both cases.

```suggestion
        when Tuple(Ast::Node, Array(Int32) | Int32)
```

src/compilers/variable.cr
```diff
@@ -96,12 +96,19 @@ module Mint
...
             when Ast::TupleDestructuring
-              js.variable_of(target.parameters[entity[1]])
+              case val = entity[1]
+              when Int32
+                js.variable_of(target.parameters[val])
+              when Array(Int32)
+                js.variable_of(val.reduce(target) do |curr_type, curr_val|
+                  curr_type.as(Ast::TupleDestructuring).parameters[curr_val]
+                end)
+              end.not_nil!
```
gdotdesign, line 111: @Sija which is more idiomatic? 

```
end.not_nil!
``` 

or

```
else
  raise "Cannot Happen!"
end
```

src/errors/error.cr
```diff
@@ -4,7 +4,7 @@ module Mint
   class Error < Exception
     alias Value = String | Ast::Node | TypeChecker::Checkable |
```
gdotdesign, line 5: Like before `Tuple(Ast::Node, Array(Int32) | Int32)` is enough here since it's matching both cases.

src/message.cr
```diff
@@ -176,7 +176,8 @@ module Mint
                                            TypeChecker::Checkable |
                                            Array(TypeChecker::Checkable) |
                                            Array(String) |
-                                           Tuple(Ast::Node, Int32))
+                                           Tuple(Ast::Node, Int32) |
```
gdotdesign, line 179: Ditto  `Tuple(Ast::Node, Array(Int32) | Int32)`

src/type_checkers/scope.cr
```diff
@@ -12,8 +12,8 @@ module Mint
                    Ast::Store |
                    Ast::Style
 
-      alias Level = Tuple(Ast::Node | Checkable | Tuple(Ast::Node, Int32), Node)
-      alias Lookup = Tuple(Ast::Node | Checkable | Tuple(Ast::Node, Int32), Node, Array(Node))
+      alias Level = Tuple(Ast::Node | Checkable | Tuple(Ast::Node, Int32) | Tuple(Ast::Node, Array(Int32)), Node)
+      alias Lookup = Tuple(Ast::Node | Checkable | Tuple(Ast::Node, Int32) | Tuple(Ast::Node, Array(Int32)), Node, Array(Node))
```
gdotdesign, line 16: Ditto  `Tuple(Ast::Node, Array(Int32) | Int32)`.

src/type_checkers/statement.cr
```diff
@@ -6,24 +6,31 @@ module Mint
...
+      } if target.parameters.size > condition.parameters.size
+
+      target.parameters.each_with_index do |param, index|
+        subcondition = condition.parameters[index]
+        _check_statement_target(param, node, subcondition)
       end
+    end
 
-      types[node] = type
+    private def _check_statement_target(target, node, condition)
```
gdotdesign, line 33: Is this needed for a `nil` branch?

src/type_checkers/variable.cr
```diff
@@ -47,7 +47,7 @@ module Mint
               type
             end
           end
-        when Tuple(Ast::Node, Int32)
+        when Tuple(Ast::Node, Int32), Tuple(Ast::Node, Array(Int32))
```
gdotdesign, line 50: Ditto `Tuple(Ast::Node, Array(Int32) | Int32)`

gdotdesign (commented):
There are some minor issues but other than that it's LGTM :+1:

src/compilers/variable.cr
Sija: In cases like these - in which I'm pretty sure the value is there - I'd use `not_nil!` since the stacktrace should be enough (except for cases in which we build mint with `--no-debug` flag - but then we're covered as long as the bug is reproducible).

gdotdesign: Got it thanks :+1:

src/parsers/tuple_destructuring.cr
```diff
@@ -31,5 +31,9 @@ module Mint
           input: data)
       end
     end
+
+    private def tuple_destructuring_value : Ast::Node?
```
Sija, line 35: I'd call it in the same vein as the other methods, akin to `tuple_destructuring_or_variable`, along with proper `Ast` classes - no casts plz.

src/type_checkers/statement.cr
matthewmcgarvey: No, there are multiple types that the where statement target could have but the only one of them we validate is tuple destructuring and that check is also recursive. Rather than do type checking if statements, I use an overload to catch and ignore anything that is not a tuple destructuring.

src/type_checkers/variable.cr
```diff
@@ -57,7 +57,12 @@ module Mint
           when Ast::Statement, Ast::WhereStatement
             case item.target
             when Ast::TupleDestructuring
-              type.parameters[value[1]]
+              case val = value[1]
+              when Int32
+                type.parameters[val]
+              when Array(Int32)
+                val.reduce(type) { |curr_type, curr_val| curr_type.parameters[curr_val] }
+              end.not_nil!
```
Sija, line 65:
```suggestion
              in Int32
                type.parameters[val]
              in Array(Int32)
                val.reduce(type) { |curr_type, curr_val| curr_type.parameters[curr_val] }
              end
```

matthewmcgarvey: Not sure how I've gone this long without knowing about this!

gdotdesign: Me neither :joy: This needs be documented here https://tw.crystal-lang.org/docs/syntax_and_semantics/case.html :D

src/compilers/variable.cr
```diff
@@ -96,12 +96,19 @@ module Mint
...
```
gdotdesign, line 111: We can use `in` here as well I guess.
```suggestion
              case val = entity[1]
              in Int32
                js.variable_of(target.parameters[val])
              in Array(Int32)
                js.variable_of(val.reduce(target) do |curr_type, curr_val|
                  curr_type.as(Ast::TupleDestructuring).parameters[curr_val]
                end)
              end
```

src/type_checkers/statement.cr
Sija, line 33:
```suggestion
    private def _check_statement_target(target : Nil, node, condition)
```

Sija (approved):
LGTM 🚀

Sija:
@matthewmcgarvey Does that supersede #283?

matthewmcgarvey:
@Sija I am not going to be touching that code but I am referencing it. It was really helpful to start with and then I found it was missing a bunch of stuff

Sija:
@matthewmcgarvey What I meant to ask was whether the #283 is obsolete after merging this PR?

matthewmcgarvey:
O, I have definitely not covered all that is covered in that PR.

Things left to do:
- Nested destructuring of enums (in that PR)
- Nested destructuring of arrays (in that PR)
- Limited nested destructuring of tuples within wheres/statements (not in that PR)

Sija:
@matthewmcgarvey Thanks, got it!

PR #439, opened 2021-05-26
Use native crystal code for PNG resizing

matthewmcgarvey:
Uses https://github.com/matthewmcgarvey/png_util for PNG resizing rather than piping out to ImageMagick.

## Performance Summary

**ImageMagick (--release):** 664ms
**PNGUtil:** 612ms
**PNGUtil (--release):** 148ms (4.5 times faster than ImageMagick!)

### DIFF
--- a/shard.lock
+++ b/shard.lock
@@ -32,6 +32,10 @@ shards:
     git: https://github.com/icyleaf/markd.git
     version: 0.4.0
 
+  png_util:
+    git: https://github.com/matthewmcgarvey/png_util.git
+    version: 0.1.0
+
   radix:
     git: https://github.com/luislavena/radix.git
     version: 0.4.1
--- a/shard.yml
+++ b/shard.yml
@@ -20,6 +20,9 @@ dependencies:
   markd:
     github: icyleaf/markd
     version: ~> 0.4.0
+  png_util:
+    github: matthewmcgarvey/png_util
+    version: ~> 0.1.0
 
 development_dependencies:
   ameba:
--- a/src/all.cr
+++ b/src/all.cr
@@ -8,6 +8,7 @@ require "uuid"
 require "html"
 require "json"
 require "xml"
+require "png_util"
 
 MINT_ENV = {} of String => String
 
--- a/src/utils/icon_generator.cr
+++ b/src/utils/icon_generator.cr
@@ -3,22 +3,12 @@ module Mint
     extend self
 
     def convert(image, size)
-      output =
-        IO::Memory.new
-
-      error =
-        IO::Memory.new
-
-      status =
-        Process.run(
-          "convert #{image} -resize #{size}x#{size} -",
-          shell: true, error: error, output: output)
-
-      if status.success?
-        output.to_s
-      else
-        ""
-      end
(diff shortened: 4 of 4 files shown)


gdotdesign:
This is great, but it only works for PNGs and ImageMagick can convert from a lot of formats to PNG so I'll merge this if we only use this for PNGs - and maybe add a flag to always use ImageMagick.

WDYT? cc @Sija

gdotdesign:
Why close it? we can still discuss it :slightly_smiling_face: I mean we can use this for PNG only since it's faster no?

Sija:
I clicked the wrong button, but as I said before I wouldn't like to introduce dependency like this for a such lil' gain in non-critical path of the app. Added complexity and maintenance is sth that IMO shouldn't be taken lightly.

matthewmcgarvey:
I did think about that as I was preparing this PR. I have a rough idea of what was making the stumpy library slower and could probably update stumpy to get similar results. It hasn't been updated in quite a while and it was made for generative art though.

About it only processing PNGs, that's all that the IconGenerator is used for so I didnt think much about handling other file types. Seems like imagemagick would be the only choice if more file types are wanted.

Performance-wise, the reason I was looking into the image processing is because I was looking at the times for`mint build`. The two high numbers were for parsing files which we fixed and for generating icons. By fixing both of those, building mint real realworld started happening in under a second

gdotdesign:
> About it only processing PNGs, that's all that the IconGenerator is used for so I didnt think much about handling other file types. Seems like imagemagick would be the only choice if more file types are wanted.

In light of that I would stay with ImageMagick for now, currently it's possible to use different source formats other than PNG but we convert them to PNG.

> Performance-wise, the reason I was looking into the image processing is because I was looking at the times for`mint build`. The two high numbers were for parsing files which we fixed and for generating icons. By fixing both of those, building mint real realworld started happening in under a second

It's awesome that you would like to optimize build times :metal: I only had issues with that because of the parsing, not much with generating icons.

Sija:
> I did think about that as I was preparing this PR. I have a rough idea of what was making the stumpy library slower and could probably update stumpy to get similar results. It hasn't been updated in quite a while and it was made for generative art though.

It's great you've taken your time to optimize the code, yet since `stumpy_png` is already used by the community, I'd advise to incorporate your optimizations upstream, instead of creating a specialized fork.

> About it only processing PNGs, that's all that the IconGenerator is used for so I didnt think much about handling other file types. Seems like imagemagick would be the only choice if more file types are wanted.

I concur with @gdotdesign. Handling the common graphic types (which boils down to `png` and `jpg` in most cases) is a feature we'd like to keep for good developer UX.

> Performance-wise, the reason I was looking into the image processing is because I was looking at the times for`mint build`. The two high numbers were for parsing files which we fixed and for generating icons. By fixing both of those, building mint real realworld started happening in under a second

I like your thinking :) I'd see this lowered down too, yet it's more a _nice to have_, than _need to have_ atm. In the end what I'd like to see is a shard (possibly binding `libvips` or other image manipulation library) with Crystal-ish API which would make image manipulation a breeze, so we could use that instead. Until then let's keep the status quo though.

PR #443, opened 2021-05-27
Show the filename and line number for each failed test

### LINKED ISSUE #441: Display the filename and the source line of the failed test
Right now the failing tests are shown with just the name (and a suite name, so that's better already), which makes navigating to them a bit hard at a times. IMO showing the filename along with the line number would be of an immense help in that regard.

Followup to #436

Sija:
Resolves #441

image

or see the [example of the failed run](https://github.com/mint-lang/mint/pull/443/checks?check_run_id=2702003093).

### DIFF
--- a/src/ast.cr
+++ b/src/ast.cr
@@ -71,7 +71,7 @@ module Mint
       node1.input.input[start_position, count].includes?('\n')
     end
 
-    def merge(ast)
+    def merge(ast) : self
       @components.concat ast.components
       @providers.concat ast.providers
       @comments.concat ast.comments
--- a/src/ast/node.cr
+++ b/src/ast/node.cr
@@ -1,6 +1,17 @@
 module Mint
   class Ast
     class Node
+      struct Location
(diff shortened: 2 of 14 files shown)


src/ast.cr
```diff
@@ -71,7 +71,7 @@ module Mint
       node1.input.input[start_position, count].includes?('\n')
     end
 
-    def merge(ast)
+    def merge(ast) : self
```
gdotdesign, line 74: Didn't know this was a thing :D

gdotdesign:
LGTM! 

The only thing is weird for me that the location is displayed at the end of the test, rspec for example shows the failures without the location and after that a failed examples group with the locations which I like more, but it's not a deal breaker for me.

rspec failures

Sija:
Well, I do see the location there, twice 🤔

Sija:
I'm gonna merge this PR as it is, and if needed we can do a followup changing the output to the desired state.

Issue #115, opened 2019-04-22
Mangle names only for final builds


TeddyDD:
Development server should not mangle names so it's easier to see Dom structure in browser's inspector

gdotdesign:
Do you mean not mangle CSS classes and variables right?

TeddyDD:
Yup, it would be great.  react-developer-tools are also helpful, It might be good idea to mention them in the docs for noobs like me :)

gdotdesign:
I've been thinking about this and the problem with using non mangled classes in development is that you can target nested components with it like this:

```
component Nested {
  style base {
    color: green;
  }

  fun render : Html {
    <div::base>
      "World"
    </div>
  }
}

component Main {
  style base {
    color: red;

    & .nested-base {
      color: yellow;
    }
  }

  fun render : Html {
    <div::base>
      "Hello"
      <Nested/>
    </div>
  }
}
```

Which would result in a yellow `World` instead of green. I myself have run into this in my project.

I'm suggesting an other way of identifying elements using the `mint-development-class` attribute which would be added on development and removed in build. I think this is less abusable then using the `class`.

I'm open to other suggestions though :)

s0kil:
What about just `{class}-mint` during development?
Long class names make it more difficult working in devtools.

justchen1369:
Perhaps append a unique identifier after the element name?

Danil-Lapirow:
> What about just `{class}-mint` during development?
> Long class names make it more difficult working in devtools.

Maybe just `{component}-{class}` because they are scoped?

gdotdesign:
What I am proposing is that elements with styles attached to them would get it as attributes (`ui-toggle--base`) this would make them easier to spot in the dev tools:

Given there is this component:
```
component Ui.Toggle {
  ...
  
  style base {
    ...
  }

  style label {
    ....
  }

  style overlay {
    ...
  }

  ...

  fun render : Html {
    <button::base
      aria-checked={Bool.toString(checked)}
      disabled={disabled}
      onClick={toggle}
      role="checkbox">

      <div::label aria-hidden="true">
        <{ onLabel }>
      </div>

      <div::label aria-hidden="true">
        <{ offLabel }>
      </div>

      <div::overlay aria-hidden="true"/>

    </button>
  }
}
```

It would seen in the dev tools as: 
```
<button 
  ui-toggle--base
  style="variables..."
  aria-checked="true"
  disabled="false"
  class="a-a">

  <div 
[...]

PR #446, opened 2021-05-28
Include css class name in compiled output

matthewmcgarvey:
Related to #115 

Instead of iterating through the alphabet, this keeps the class name and pairs it with a unique hash connected to each style's component.

Normally having a `base` class in mint would generate something like `aa` as the class when you inspect the browser. Now it will look like `_base_c1010`

This does not update the variables. I'm not entirely sure how they work right so I didn't try to figure out how they could fit into this. There's also a ton of failing tests. I haven't quite figured out how to keep the hash id consistent int tests and I wanted to get feedback on this implementation anyways.

The connected issue talks about only mangling them in the final build. I'm not too sure if that's necessary. Let me know and I'll try to figure out how that works.

### DIFF
--- a/spec/compilers/css_definition
+++ b/spec/compilers/css_definition
@@ -13,7 +13,7 @@ component Main {
 }
 --------------------------------------------------------------------------------
 class A extends _C {
-  $a() {
+  $test_HASH_ID() {
     const _ = {
       [`--a-a`]: this.a + `px 0px`
     };
@@ -27,8 +27,8 @@ class A extends _C {
 
   render() {
     return _h("div", {
-      className: `a`,
-      style: _style([this.$a()])
+      className: `test_HASH_ID`,
+      style: _style([this.$test_HASH_ID()])
     });
   }
 };
@@ -36,7 +36,7 @@ class A extends _C {
 A.displayName = "Main";
 
 _insertStyles(`
-.a {
+.test_HASH_ID {
   margin: var(--a-a);
 }
 `);
--- a/spec/compilers/css_font_face
+++ b/spec/compilers/css_font_face
@@ -20,7 +20,7 @@ component Main {
 class A extends _C {
   render() {
     return _h("div", {
-      className: `a`
+      className: `test_HASH_ID`
     });
   }
 };
--- a/spec/compilers/css_font_face_with_qoutes
+++ b/spec/compilers/css_font_face_with_qoutes
@@ -16,7 +16,7 @@ component Main {
 class A extends _C {
   render() {
     return _h("div", {
-      className: `a`
+      className: `test_HASH_ID`
     });
   }
 };
@@ -24,7 +24,7 @@ class A extends _C {
 A.displayName = "Main";
 
 _insertStyles(`
-.a {
+.test_HASH_ID {
   font-family: "myFirstFont";
 }
 
--- a/spec/compilers/css_keyframes
+++ b/spec/compilers/css_keyframes
@@ -37,7 +37,7 @@ class A extends _C {
     });
   }
 
-  $a() {
+  $test_HASH_ID() {
     const _ = {
       [`--a-a`]: this.a
     };
@@ -51,8 +51,8 @@ class A extends _C {
 
   render() {
     return _h("div", {
-      className: `a`,
-      style: _style([this.$a()])
+      className: `test_HASH_ID`,
+      style: _style([this.$test_HASH_ID()])
     });
   }
 };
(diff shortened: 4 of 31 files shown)


gdotdesign:
Thank you for working on this :pray:, in general the implementation seems right.

> The connected issue talks about only mangling them in the final build. I'm not too sure if that's necessary. Let me know and I'll try to figure out how that works.

Using the alphabet in the final build is mainly for reducing the size of the generated JS and CSS, so I think that's necessary.

matthewmcgarvey:
When creating an optimized build, it switches back to incrementing letters. I also had to make hash_id on components a property or else there would be no way to test the class names without switching to optimized mode

Sija:
Could you please revert changes to the specs? There should be a dedicated spec testing just this functionality.

matthewmcgarvey:
@Sija I think you're assuming this isn't going to be the default behavior? Is that what you mean? What kind of set up would enable this?

The specs are changed because they are run through the compiler with optimized turned off.

Sija:
Not just that. Another thing is that it makes reviewing this PR unnecessarily hard :) And yet another, is that if that's only for debugging purposes you might want to compile the code with optimizations turned on but only this setting enabled - I can imagine debugging things on staging/live env.

gdotdesign:
I've pushed my take on this here https://github.com/mint-lang/mint/commit/c9fdd5537c670c18d03e8cdd2482a6825e9bee6b

The class name of the not optimized version is now like this: 
```
Ui·Dropdown_base_DO_NOT_USE_MINIMIZED_IN_BUILD_CODE
```

1. Name of the component where dots `.` are replaced with `·` - starts with uppercase 
2. Name of the style - which starts with lowercase
3. `DO_NOT_USE_MINIMIZED_IN_BUILD_CODE` I know this makes the HTML verbose but I think this is necessary because if the class is remotely readable people will use it to target the classes of sub components. 

With this there can be no conflict between the styles of the components, even when name spaced.

------ 

I don't know which should we test by default optimized or not optimized version I would go with not optimized because for tests it's more readable.

-----------

@matthewmcgarvey @Sija let me know what you think, if it's good I can merge my commit into this PR.

Sija:
I'd change a couple of things to make reviewing this a bearable experience :)

1. Reverting all changes to the tests, eliminating needless diff
2. Using an HTML `data` attribute, sth like a `data-mint-style` holding the value with no replacements being made, and no `_DANGER` suffix
3. Add the above attribute only in cases where `optimize` is `false`

gdotdesign:
> Reverting all changes to the tests, eliminating needless diff

As I wrote before, I don't know which should we test by default: the optimized or not optimized version I would go with not optimized because for tests it's more readable that means the diffs would stay.

> Using an HTML `data` attribute, sth like a `data-mint-style` holding the value with no replacements being made, and no `_DANGER` suffix - while you can still target it via CSS selectors, it's less likely that someone will abuse it that _easily_ - and sometimes you might _want to_ abuse it for purpose ;)

My original idea was to use the an attribute, but the main issue is that it would be there in development (not optimized) but not there in production (optimized) so if anyone abuses that it **will break unexpectedly** so we need to let them know not to do this.

matthewmcgarvey:
Just for reference, I did try switching the tests to the optimized version to avoid having to fix the CSS classes and it produces waaay more and bigger differences/failures.

I like what @gdotdesign did with creating a StylePool. Maybe we can make that an abstract class and have an OptimizedStylePool and a DevelopmentStylePool that we can choose and pass into the StyleBuilder. That way the tests could use the OptimizedStylePool and not have to be changed (or changed in a later PR) and the DevelopmentStylePool could be tested independently. I'm not that big of a fan of having that `_DANGER` suffix instead of the hash because I think the hash makes it impossible rather than just warning (plus it's only 6 letters extra).

gdotdesign:
>  I'm not that big of a fan of having that `_DANGER` suffix instead of the hash because I think the hash makes it impossible rather than just warning (plus it's only 6 letters extra).

The hash is created from the name so it won't change much which can be used to target it. 

Without the `_DANGER` suffix I don't know how will someone know that they should not target that element with the class. The only other way would be to inject 
```
console.warn("!!!DON'T TARGET ELEMENTS WITH SELECTORS BECAUSE THE SELECTORS WILL BE MINIMIZED IN THE PRODUCTION BUILD!!!")
```

in development mode but that too can be easily ignored.

matthewmcgarvey:
@gdotdesign @Sija feel free to take over this PR. I don't have as much time as I hoped to think through everything

Sija:
Superseded by #467

PR #451, opened 2021-05-31
Add Inline Comment

### LINKED ISSUE #81: Consider line comment

wisn:
## Summary

Add inline comment support.
This fixes #81.


src/ast/comment.cr
```diff
@@ -1,9 +1,10 @@
 module Mint
   class Ast
     class Comment < Node
-      getter value
+      getter value, type
```
gdotdesign, line 4: I would use an enum for this:

```
enum CommentType
  Inline
  Block
end
```

gdotdesign (approved):

src/ast/comment.cr
wisn: @gdotdesign sure

src/ast/comment.cr
```diff
@@ -1,9 +1,15 @@
 module Mint
   class Ast
+    enum CommentType
+      Inline
+      Block
+    end
```
Sija, line 6: Please move into a `Comment::Type`.

src/ast/comment.cr
```diff
@@ -1,9 +1,15 @@
...
+
     class Comment < Node
-      getter value
+      getter value, type
 
       def initialize(@value : String,
+                     @type : CommentType,
```
Sija, line 12:
```suggestion
                     @type : Type,
```

src/formatters/comment.cr
```diff
@@ -7,10 +7,18 @@ module Mint
           .remove_leading_whitespace
           .rstrip
 
-      if replace_skipped(value).includes?('\n')
-        "/*\n#{value}\n*/"
+      if node.type == Ast::CommentType::Block
```
Sija, line 10:
```suggestion
      if node.type.block?
```

src/parsers/comment.cr
```diff
@@ -2,17 +2,32 @@ module Mint
...
+            gather { consume_while((!(keyword_ahead "*/") || char == '\0') && !eof?) }.to_s
+
+          keyword "*/"
+        else
+          keyword "//"
+
+          value =
+            gather { consume_while(!((keyword_ahead "\n") || char == '\0') && !eof?) }.to_s
+          type = Ast::CommentType::Inline
+        end
```
Sija, line 23:
```suggestion
        value = nil

        if keyword_ahead "/*"
          keyword "/*"

          type = Ast::CommentType::Block
          value =
            gather { consume_while((!(keyword_ahead "*/") || char == '\0') && !eof?) }.to_s

          keyword "*/"
        else
          keyword "//"

          type = Ast::CommentType::Inline
          value =
            gather { consume_while(!((keyword_ahead "\n") || char == '\0') && !eof?) }.to_s
        end
```

Sija (changes requested):

Sija (approved):
LGTM, thanks @wisn! 🚀

PR #453, opened 2021-06-01
Add timestamp to AnimationFrame

### LINKED ISSUE #445: Provider.AnimationFrame does not provide DOMHighResTimeStamp argument
Hi! 👋 Wow, Mint has grown a lot since I last used it. The home page is beautiful and the documentation is excellent. Great work! 

While looking at built-in solutions for `requestAnimationFrame` I came across the `AnimationFrame` provider. Unfortunately, it does not provide components with the [`DOMHighResTimeStamp` argument](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame#parameters) that `requestAnimationFrame` sends to the callbacks, which prevents deriving the elapsed time since the last animation frame. I was able to make my own provider to do this, but I thought this was worth reporting.

bbugh:
This adds the expected `Number` argument from `requestAnimationFrame` so that consumers can calculate the time step difference, etc.

Fixes #445

### DIFF
--- a/core/source/Test/Context.mint
+++ b/core/source/Test/Context.mint
@@ -6,7 +6,7 @@ module Test.Context {
     test {
       with Test.Context {
         of(5)
-        |> Test.assertEqual(5)
+        |> assertEqual(5)
       }
     }
   */
@@ -20,7 +20,7 @@ module Test.Context {
     test {
       with Test.Context {
         of(5)
-        |> then(\number : Number => Promise.resolve(number + 2))
+        |> then((number : Number) { Promise.resolve(number + 2) })
         |> assertEqual(7)
       }
     }
@@ -36,7 +36,17 @@ module Test.Context {
     `
   }
 
-  /* Adds a timeout to the text using the given duration (in milliseconds). */
+  /*
+  Adds a timeout to the text using the given duration (in milliseconds).
+
+    test {
+      with Test.Context {
+        of(5)
+        |> timeout(1000)
+        |> assertEqual(5)
+      }
+    }
+  */
   fun timeout (duration : Number, context : Test.Context(a)) : Test.Context(a) {
     then(
       (subject : a) : Promise(Never, a) { Timer.timeout(duration, subject) },
@@ -49,24 +59,33 @@ module Test.Context {
     test {
       with Test.Context {
         of(5)
-        |> Test.assertEqual(5)
+        |> assertEqual(5)
       }
     }
   */
-  fun assertEqual (a : a, context : Test.Context(a)) : Test.Context(a) {
+  fun assertEqual (value : a, context : Test.Context(a)) : Test.Context(a) {
     `
     #{context}.step((subject)=> {
-      let result = _compare(#{a}, subject)
+      let result = _compare(#{value}, subject)
 
       if (result) {
         return subject
       } else {
-        throw \`Assertion failed ${#{a}} === ${subject}\`
+        throw \`Assertion failed ${#{value}} === ${subject}\`
       }
     })
     `
   }
 
+  /*
+  Asserts if the given value equals of the returned value from the given
+  function.
+
+    with Test.Context {
+      of(5)
+      |> assertOf("5", Number.toString)
+    }
+  */
   fun assertOf (
     value : b,
     method : Function(a, b),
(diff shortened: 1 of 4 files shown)


gdotdesign:
Please update the tests, otherwise it looks good :+1:

bbugh:
Ah, I forgot to mention that. Hard to think while holding a squirmy child... 

I wasn't sure how to test the timestamps, given that the argument is non-deterministic. In other tools, I'd use a mock or stub tool, or something like [timecop](https://github.com/travisjeffery/timecop) to guarantee consistent results.

I looked through the source for some way to mock it and/or make a strict timestamp result, and I couldn't find anything. Do you have a suggestion?

Sija:
@bbugh I'd change tests to assert that the passed argument is greater than `0`. That should be good enough.

bbugh:
There does not seem to be an assertion for greater than, and I don't see one for not equal, or any negation at all. I will make another issue for this.

bbugh:
Added discussion https://github.com/mint-lang/mint/issues/458

core/source/Test/Context.mint
```diff
@@ -36,7 +36,17 @@ module Test.Context {
     `
   }
 
-  /* Adds a timeout to the text using the given duration (in milliseconds). */
+  /*
+  Adds a timeout to the text using the given duration (in milliseconds).
```
Sija, line 40:
```suggestion
  Adds a timeout to the test using the given duration (in milliseconds).
```

core/source/Test/Context.mint
```diff
@@ -49,21 +59,33 @@ module Test.Context {
...
-        throw \`Assertion failed: ${#{a}} === ${subject}\`
+    #{context}.step((subject)=> {
+      let result = _compare(#{value}, subject)
+
+      if (result) {
+        return subject
+      } else {
+        throw \`Assertion failed ${#{value}} === ${subject}\`
       }
-      return subject
```
Sija, line 62: Please, put all of the changes to the `Context` module into a different PR instead.

core/tests/tests/Provider/AnimationFrame.mint
```diff
@@ -1,23 +1,30 @@
 component Test.Provider.AnimationFrame {
-  state frames : Number = 0
+  state timestamp : Number = 0
 
-  use Provider.AnimationFrame { frames = () : Promise(Never, Void) { next { frames = frames + 1 } } }
+  use Provider.AnimationFrame {
+    frames =
+      (timestamp : Number) : Promise(Never, Void) {
+        next { timestamp = timestamp }
+      }
```
Sija, line 8:
```suggestion
    timestamp =
      (timestamp : Number) : Promise(Never, Void) {
        next { timestamp = timestamp }
      }
```

Sija (changes requested):

bbugh:
@gdotdesign thanks for the updates! It's useful to see how you would write those tests, as the resident Mint expert. ;) 

@Sija in case you didn't notice, the changes you've requested were put in by @gdotdesign's commits, so I'm unsure if I should accept the suggestions.

core/tests/tests/Provider/AnimationFrame.mint
```diff
@@ -1,23 +1,30 @@
...
-      |> assertTextOf("div", "1")
-      |> assertTextOf("div", "2")
-      |> assertTextOf("div", "3")
+    with Test.Context {
+      with Test.Html {
+        <Test.Provider.AnimationFrame/>
+          |> start()
+          |> find("div")
+          |> map(Dom.getTextContent)
+          |> assertNotEqual("0")
```
Sija, line 26: We shouldn't need `assertNotEqual `:

```suggestion
        (<Test.Provider.AnimationFrame/>
          |> start()
          |> find("div")
          |> map(Dom.getTextContent)) != "0"
```

core/source/Test/Context.mint
Sija: I've opened up #460 to address this.

core/tests/tests/Provider/AnimationFrame.mint
Sija: Btw, I don't think we even need to test the argument in any way ;)
We ain't responsible for its lifecycle, nor validity, so... 🤷‍♂️

gdotdesign: This wouldn't work since the left side is a `Test.Context(String)` and the right side is a `String`.

How tests work is that either it returns:

* `Bool` - if its `true` the test passes and if it's `false` the tests fails
* `Test.Context(a)` - a multi-step test, the steps are basically promises if step (promise) fails then the whole test fails, if all resolves then the test passes.

PR #454, opened 2021-06-02
Add `Maybe.withLazyDefault`

Sija:


gdotdesign (approved):

Sija:
@gdotdesign I don't like the name, I couldn't come up with anything better though. If function overloading was a thing I'd add it as another overload of `Maybe.withDefault`...

gdotdesign:
> @gdotdesign I don't like the name, I couldn't come up with anything better though.

Probably using **lazy** as in `Maybe.withLazyDefault` - probably that's what FP people are familiar with. The other option I can think of to be using **deferred** - `Maybe.withDeferredDefault`.

Sija:
I like `Maybe.withLazyDefault` 👍

PR #467, opened 2021-06-18
Include css class name in compiled output

### LINKED ISSUE #115: Mangle names only for final builds

gdotdesign:
Supersedes #446 with original commits intact and resolves #115


matthewmcgarvey (approved):

src/style_builder.cr
```diff
@@ -192,14 +214,14 @@ module Mint
...
 
-    def prefixed_class_name(node)
-      @css_prefix.to_s + style_pool.of(node, nil)
+    def prefixed_class_name(node : Ast::Style, id : String? = nil)
+      @css_prefix.to_s + style_pool.of(node, id)
     end
 
     # The main entry point for processing a "style" tag.
-    def process(node : Ast::Style)
+    def process(node : Ast::Style, id : String)
```
Sija, line 222: Shouldn't this allow for nil-able `id` too?

```suggestion
    def process(node : Ast::Style, id : String? = nil)
```

src/style_builder.cr
```diff
@@ -17,6 +17,28 @@ module Mint
...
+
+    def of(subject : Ast::Style, id : String? = nil)
+      if @optimize
+        @pool.of(subject, nil)
+      else
+        @cache[subject] ||= begin
+          if id
+            "#{id}_#{subject.name.value.gsub('-', '_')}"
+          else
+            subject.name.value
```
Sija, line 35: Shouldn't this be escaped too?

```suggestion
            subject.name.value.gsub('-', '_')
```

gdotdesign: I don't think so, why?

Sija: In the same vein as the line above I guess.

gdotdesign: Oh right, I was thinking you commented on *that* line :joy:

src/style_builder.cr
gdotdesign: It compiles so _I guess not_?

Sija: I'm asking since the API all around it allows for empty `id`, so I don't see a reason why this one shouldn't.

Sija: It calls `prefixed_class_name` which has nil-able `id`, this, in turn, invokes `StylePool#of` which does the same, etc.

Sija: > It compiles so _I guess not_?

Just because there are no tests stressing this path :P

PR #472, opened 2021-07-03
Fetch resource over the network before falling back to `index.html`

### LINKED ISSUE #370: Incorrect fetching from service-worker
When the service-worker doesn't find a cache for a file, it responds with the file `index.html`.

The problem is that **it doesn't check if the requested file is one of the cacheable files**.

If not, it should fetch the request instead of serving from the cache.

Sija:
Fixes #370

### DIFF
--- a/src/utils/service_worker.ecr
+++ b/src/utils/service_worker.ecr
@@ -46,11 +46,18 @@ self.addEventListener('fetch', event => {
   // If we are on the same origin
   if (isSameOrigin) {
     // Try to get the response from the cache, and if not available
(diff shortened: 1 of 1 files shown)


Sija:
One thing to consider is long-running requests, which can hang for as long as 300 seconds (Chrome) - for more information see https://dmitripavlutin.com/timeout-fetch-request/.

I'm not sure if that poses any significant risk to the overall functionality, but I thought it's sth worth mentioning.

Sija:
Since this will result in issuing an HTTP request for every navigational request, it feels like a wrong approach altogether.

PR #476, opened 2021-07-18
Feat: allow to swap runtime at compilation

WhoAteDaCake:


src/builder.cr
```diff
@@ -128,9 +128,15 @@ module Mint
...
     end
 
-    def index(css_prefix, relative, optimize)
+    def index(css_prefix, relative, optimize, runtime_path)
       runtime =
-        Assets.read("runtime.js")
+        if runtime_path
+          File.open(runtime_path) do |file|
+            file.gets_to_end
+          end
```
Sija, line 136:
```suggestion
          File.read(runtime_path)
```

src/commands/compile.cr
```diff
@@ -16,18 +16,28 @@ module Mint
...
+      def compile(optimize, runtime_path)
         json =
           MintJson.parse_current
 
         runtime =
-          Assets.read("runtime.js")
+          if runtime_path
+            File.open(runtime_path) do |file|
+              file.gets_to_end
+            end
```
Sija, line 37:
```suggestion
            File.read(runtime_path)
```

Sija (changes requested):

src/builder.cr
```diff
@@ -128,9 +128,13 @@ module Mint
       end
     end
 
-    def index(css_prefix, relative, optimize)
+    def index(css_prefix, relative, optimize, runtime_path)
       runtime =
-        Assets.read("runtime.js")
+        if runtime_path
```
gdotdesign, line 133: Maybe we should check if the specified file exists? This way it will raise an exception if the file is not found - ditto on the compile command as well.

It's not blocking this PR just would be nice to have :slightly_smiling_face:

gdotdesign (approved):

src/commands/build.cr
```diff
@@ -23,13 +23,18 @@ module Mint
         default: true,
         short: "m"
 
+      define_flag runtime : String,
+        description: "Will use supplied runtime instead of the default distribution",
```
Sija, line 27:
```suggestion
        description: "Will use supplied runtime path instead of the default distribution",
```

src/builder.cr
WhoAteDaCake: What type of exception would you expect to be thrown? Also, If I'm adding a bit for complex logic, would it make sense to extract this to a module, so I don't duplicate it ?

WhoAteDaCake: `File::NotFoundError` should be thrown if the file doesn't exist by the `File.read`. Should we use something like `print` to stderr and exit with 1, rather than having crystal print all stack trace with it?

gdotdesign: > Should we use something like print to stderr and exit with 1, rather than having crystal print all stack trace with it?

Yeah, that's what I was thinking of.

> Also, If I'm adding a bit for complex logic, would it make sense to extract this to a module, so I don't duplicate it ?

Ideally yes, but I'm not sure two occurrences warrants it or not, @Sija what do you think?

Sija: > Ideally yes, but I'm not sure two occurrences warrants it or not, @Sija what do you think?

I don't think so, nope. It gives almost no gain considering it saves a few lines, and creates additional abstraction and a dependency.

Sija: > What type of exception would you expect to be thrown?

@WhoAteDaCake I'd suggest creating sth like `RuntimeFileNotFound` `message` (akin to `EnvFileNotFound`) and register it as a `command_error`.

WhoAteDaCake: I've released that we're also using `runtime.js` within `sandbox_server.cr` and `test_runner.cr`. Should both of those also allow for swapping the runtime? If so, could you provide me a bit of guidance for the approach ?

src/messages/runtime_not_found.cr
```diff
@@ -0,0 +1,9 @@
+message RuntimeFileNotFound do
+  title "RuntimeFileNotFound Error"
+
+  block do
+    text "The specified runtime path"
```
Sija, line 5:
```suggestion
    text "The specified runtime file"
```

src/builder.cr
Sija: I'd think so. /cc @gdotdesign

src/messages/runtime_not_found.cr
```diff
@@ -0,0 +1,9 @@
+message RuntimeFileNotFound do
+  title "RuntimeFileNotFound Error"
```
Sija, line 2:
```suggestion
  title "Runtime Error"
```

src/messages/runtime_not_found.cr
```diff
@@ -0,0 +1,9 @@
+message RuntimeFileNotFound do
```
Sija, line 1: Please, rename this file according to the message name, i.e.: `runtime_file_not_found.cr`

Sija: No worries :)

src/sandbox_server.cr
```diff
@@ -82,6 +82,16 @@ module Mint
       artifacts =
         TypeChecker.check(ast)
 
+      runtime =
```
WhoAteDaCake, line 85: Is it better to load the runtime once and just store it in memory? I imagine in a sandbox server restarts and loading of the runtime happens quite often

src/sandbox_server.cr
```diff
@@ -38,7 +38,7 @@ module Mint
       end
     end
 
-    def initialize(@host = "0.0.0.0", @port = 8080)
+    def initialize(@host = "0.0.0.0", @port = 8080, @runtime_path : String | Nil = nil)
```
Sija, line 41:
```suggestion
    def initialize(@host = "0.0.0.0", @port = 8080, @runtime_path : String? = nil)
```

src/sandbox_server.cr
Sija: yup, it should be enough to load it just once within the `initialize`.

src/sandbox_server.cr
WhoAteDaCake: Thanks for that, it's cleaner !

src/test_runner.cr
```diff
@@ -209,7 +209,14 @@ module Mint
...
       get "/runtime.js" do
-        Assets.read("runtime.js")
+        if runtime_path = @flags.runtime
+          raise RuntimeFileNotFound, {
+            "path" => runtime_path,
+          } unless ::File.exists?(runtime_path)
+          ::File.read(runtime_path)
+        else
+          Assets.read("runtime.js")
+        end
```
Sija, line 219: This could be loaded into memory as well, preferably within the local variable after the `page_source` definition.

WhoAteDaCake: Done :)

WhoAteDaCake:
Is that good enough to merge? Or could you approve to run against workflow for validation,

Sija (approved):
LGTM ✨

Sija:
Thanks @WhoAteDaCake! 🙏

WhoAteDaCake:
> Thanks @WhoAteDaCake! 🙏

Thank you for your patience

PR #493, opened 2021-09-29
Add default values to LSP client capabilities

esensar:
Keys in client capabilities are optional. This change makes sure that
LSP server does not fail when keys are not present (which should have
same meaning as a false value).


Sija:
Thanks, @esensar! ❤️

PR #520, opened 2021-11-26
Implement support for disabling snippets (LS)

### LINKED ISSUE #466: LSP: Support for disabling snippets
My LSP client has snippet support disabled, but they still appear as suggestions in completion.

gdotdesign:
Fixes #466


Sija:
@gdotdesign Anything left to do here before it's GTG?

gdotdesign:
> @gdotdesign Anything left to do here before it's GTG?

Just testing it with an actual client where it's not supported.

Issue #521, opened 2021-11-30
Time intervals/deltas and transformations.


frnco:
While working with data that expires I notices it's quite hard to get a `Time` that is relative to `Time.now`, and after searching the issues I wasn't able to find any proposals regarding this, except for [137](https://github.com/mint-lang/mint/issues/137) which actually gave me some ideas on a simple solution to make this work.

My proposal is fairly simple, but first I'll list some options I've considered. The first was adding stuff inline with `nextWeek`, stuff like `tomorrow` etc, but that's still very limited so I went on to consider adding Arithmethical/Transformation functions, i.e. `Time.now |> Time.addDays(2)` or `Time.now |> Time.addMinutes(30)`. But since we already have Unix, it seems pretty easy to just have something like `(Time.toUnixTimestamp(TIme.now) + ( 30 * 60 * 1000 )) |> Time.fromUnixTimestamp`, here considering the `UnixTimestamp` thing in Mint actually takes milliseconds into account, though I believe it'd be nice to include some functions into `Time` to allow us to explicitly make that distinction.

The change would actually be limited to the `TIme` module, and would consist of simply adding `toSomeEpochFormat` and `fromSomeEpochFormat` functions, which would allow anyone to modify time using simple arithmetic.

Sure, it'd be cool to have something like Ruby available, i.e. `TIme.now |> Time.add(Time.days(3))` but unfortunately `Time` does not deal with deltas. I did start to work on a `Timespan` module to deal with that but I'm nowhere near happy with how I went about doing it (A big factor was my failure to figure out how to make it a custom type consisting of basically a single number, which corresponds to the number of milliseconds for that timespan/timedelta. Instead I had to use Javascript to convert it to and from `Number`, which completely screwed up the elegant approach I came up with at the beginning and made me put that idea aside, at least for now). I wanna get back to that but first I need to figure out a way to make it a custom type that I can do arithmetic with and convert to and from Numbers without having to use inline Javascript, but the only way I can see that happening means changing the compiler, so it's postponed until I figured out a way to deal with the compiler (Or how to make the compiler friendlier), unfortunately.

Any suggestions on how to make this easier are appreciated, but for now the easiest option I can see is changing the `Time` module to painlessly convert to and from UNIX/POSIX Timestamps/Epochs, which would allow anyone to use `Number`s' to do basic math to modify the time and then convert it back. I can make the change myself and send a PR with the modified `core/source/Time.mint`, basically it'd be something like:
```
[...]

PR #522, opened 2021-12-02
Time conversions from and to UNIX Timespamps, plus a small typo fix.

frnco:
Based on the conversation #521 this PR introduces methods that convert between Unix Timestamps and an object/instance of the Javascript `Date` class to the `Time` module. Since the `Time` type is basically just an object/instance of the Javascript `Date` class, this effectively converts between `Time` and an UNIX Timestamp.

It was also pointed out that ideally UTC should be used as much as possible, and after looking into Javascript's `Date` class it appears that `setTime` and `getTime` don't make any assumptions regarding Timezones and that loading a Timestamp won't change the Timezone, which felt like the most clear way to do this.

Tests use the standard "April 5th, 2018" date used in other tests, and the code was based off of the code already used in the `Time` module, especially `now` and `nextWeek` which were used as a reference for how to initialize a `Date` and load a `Time` at first, though in the case of `toUnixTimestampInMs` it was later dropped in favor of simply using `#{time}.getTime()`.

### DIFF
--- a/core/source/Time.mint
+++ b/core/source/Time.mint
@@ -51,7 +51,7 @@ module Time {
   }
 
   /*
-  Retruns a new UTC date from the given parameters.
+  Returns a new UTC date from the given parameters.
 
     Time.from(2018, 4, 5)
   */
@@ -187,5 +187,21 @@ module Time {
       return DateFNS.distanceInWordsStrict(#{now}, #{other}, { addSuffix: true })
     })()
     `
+  }
+
+  /* Returns the time respective to the given UNIX Timestamp (in Milliseconds) */
+  fun fromUnixTimestampInMs (timestamp : Number) : Time {
+  `
+    (() => {
+      const time = new Date()
+      time.setTime(#{timestamp})
+      return time
+    })()
+    `
+  }
+
+  /* Returns the UNIX Timestamp (in Milliseconds) respective to the given time */
+  fun toUnixTimestampInMs (time : Time) : Number {
+    `#{time}.getTime()`
   }
 }
--- a/core/source/Time.mint
+++ b/core/source/Time.mint
@@ -191,7 +191,7 @@ module Time {
 
   /* Returns the time respective to the given UNIX Timestamp (in Milliseconds) */
   fun fromUnixTimestampInMs (timestamp : Number) : Time {
-  `
+    `
     (() => {
       return new Date(#{timestamp})
     })()
--- a/core/source/Time.mint
+++ b/core/source/Time.mint
@@ -193,9 +193,7 @@ module Time {
   fun fromUnixTimestampInMs (timestamp : Number) : Time {
   `
     (() => {
-      const time = new Date()
-      time.setTime(#{timestamp})
-      return time
+      return new Date(#{timestamp})
     })()
     `
   }
--- a/core/tests/tests/Time.mint
+++ b/core/tests/tests/Time.mint
@@ -74,3 +74,17 @@ suite "Time.nextWeek" {
     |> Time.toIso()) == "2018-04-12T00:00:00.000Z"
   }
 }
+
+suite "Time.fromUnixTimestampInMs" {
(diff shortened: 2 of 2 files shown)


core/tests/tests/Time.mint
```diff
@@ -74,3 +74,17 @@ suite "Time.nextWeek" {
     |> Time.toIso()) == "2018-04-12T00:00:00.000Z"
   }
 }
+
+suite "Time.fromUnixTimestampInMs" {
+  test "returns the UTC time respective to the given UNIX Timestamp (in Milliseconds" {
```
Sija, line 79:
```suggestion
  test "returns the UTC time from the given UNIX timestamp (in milliseconds" {
```

Sija (commented):
LGTM overall, left few small comments.

core/source/Time.mint
```diff
@@ -188,4 +188,20 @@ module Time {
     })()
     `
   }
+
+  /* Returns the time respective to the given UNIX Timestamp (in Milliseconds) */
+  fun fromUnixTimestampInMs (timestamp : Number) : Time {
+  `
+    (() => {
```
gdotdesign, line 195: We can just use `new Date(timestamp)` which results in the same `Date`.

```
 new Date(1522886400000).toISOString()
 // "2018-04-05T00:00:00.000Z"
 ```

**Review by gdotdesign — APPROVED** [review]

**Comment by frnco** [comment]
I've abridged the test names as @Sija suggested, and changed the code as suggested by @gdotdesign . I didn't however figure out why the checks had failed, if I could get some pointers on how to run those checks in my machine (I do have Crystal installed, I'm just not familiar with the specific processes used for running the Crystal tests for Mint, so if anyone can tell me which commands to run, that's enough and I can take it from there).

**Comment by gdotdesign** [comment]
You can check the workflow (https://github.com/mint-lang/mint/blob/master/.github/workflows/ci.yml#L54) which command fails, in this case the check of the formatting of the core files fail.

Once you build Mint from source using `make build` it will put the resulting binary in  the `bin` directory. Using that you can format the core files:

```
# Formatting core
cd core
../bin/mint format

# Formatting tests
cd core/tests
../../bin/mint format
```

frnco:
> You can check the workflow (https://github.com/mint-lang/mint/blob/master/.github/workflows/ci.yml#L54) which command fails, in this case the check of the formatting of the core files fail.
> 
> Once you build Mint from source using `make build` it will put the resulting binary in the `bin` directory. Using that you can format the core files:
> 
> ```
> # Formatting core
> cd core
> ../bin/mint format
> 
> # Formatting tests
> cd core/tests
> ../../bin/mint format
> ```

That did it. Formatted the files and pushed the changes, let's hope everything passes. Still, the formatter is quite nitpicky especially considering Mint relies on `{`braces`}` as opposed to something like Python which relies on indentation. With this level of care I actually find it weird that Mint uses braces instead of Crystal's (And Ruby's) more readable `end`.


Ranting aside, after formatting all the tests in the checks passed, and looks like the build will finish without problems too. All the proposed changes have also been implemented, so it'd be good to get further feedback on whether or not there's something else that I should add.


Plus, I made some further suggestions about adding some small extra functionality to `Time` module on #521 which could use some feedback from someone on the core team (@gdotdesign was the one I was talking to but everyone's welcome to join).

gdotdesign:
> With this level of care I actually find it weird that Mint uses braces instead of Crystal's (And Ruby's) more readable `end`.

Mint uses braces because it wants to be similar looking to JavaScript, so JavaScript devs wouldn't be put off using it.



> Ranting aside, after formatting all the tests in the checks passed, and looks like the build will finish without problems too. All the proposed changes have also been implemented, so it'd be good to get further feedback on whether or not there's something else that I should add.

It's good to go :+1: Thank you :pray: 

> Plus, I made some further suggestions about adding some small extra functionality to `Time` module on #521 which could use some feedback from someone on the core team (@gdotdesign was the one I was talking to but everyone's welcome to join).

Thank you for that too :pray: I'll check it tomorrow.

PR #524, opened 2021-12-05
Update Clipboard.set to restore clobbered global state

ryanprior:
The `Clipboard.set` method clobbers the global state, removing focus from the current active element and removing all content selections. This PR updates the method to remember the initial active element and selection(s), restoring them after the clipboard operation is complete.

It also updates `Clipboard.set` to return its argument for convenient chaining and removes `;` which was put at the end of some lines but not others, and isn't necessary.

### DIFF
--- a/core/source/Clipboard.mint
+++ b/core/source/Clipboard.mint
@@ -1,57 +1,56 @@
 /* This module has functions for manipulating the clipboard. */
 module Clipboard {
   /* Sets the clipboards content to the given value. */
-  fun set (value : String) : String {
-    try {
-      `
-      (() => {
-        // Create a textarea element
-        const textarea = document.createElement("textarea")
+  fun set (value : String) : Promise(Never, String) {
+    `
+    (() => {
+      // Create a textarea element
+      const textarea = document.createElement("textarea")
 
-        // Position it on the screen
-        textarea.style.position = "fixed"
-        textarea.style.left = "10px"
-        textarea.style.top = "10px"
-        textarea.style.opacity = 0
+      // Position it on the screen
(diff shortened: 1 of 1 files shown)


core/source/Clipboard.mint
```diff
@@ -1,43 +1,57 @@
...
-      textarea.setSelectionRange(0, 999999);
-
-      // Copy to clipboard
-      document.execCommand("copy");
-
-      // Remove textarea from the DOM
-      textarea.remove()
-    })()
-    `
+  fun set (value : String) : String {
```
gdotdesign, line 4: I would keep it as a `Promise(Never, String)` mainly because at some point it will support the [async clipboard api](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API).

core/source/Clipboard.mint
```diff
@@ -1,43 +1,57 @@
...
+
+        // Refocus the previous active element
+        lastActive.focus()
+
+        // Restore previous range(s)
+        selection.removeAllRanges()
+        for(range of lastRanges) selection.addRange(range)
+      })()
+      `
+      value
```
gdotdesign, line 54: No need to wrap it in a  `try` it can be just `return #{value}`.

core/source/Clipboard.mint
ryanprior: Would you like me to go ahead and rewrite to to use that API?

gdotdesign: If you feel like it, that would be great :) In another PR and using this as a fallback (`if ('clipboard' in navigator) ...`).

gdotdesign (approved):

PR #526, opened 2021-12-13
Implement Mint native `Time` functions.

gdotdesign:
This PR implements `Time` related functions  in Mint instead of relying on a library. 

For a long time I wanted to do this change. Thanks to @frnco for creating issues related to `Time` (https://github.com/mint-lang/mint/issues/521) which got me excited and pushed me to do this finally :)

- Drops dependency on [DateFNS](https://github.com/mint-lang/mint-runtime/blob/master/package.json#L18).
- Implements everything that was supported before (with breaking changes)
- A lot of inspiration was taken from [Crystals Time Module](https://crystal-lang.org/api/1.2.2/Time.html) (and `calendar_week` was pretty much ported over)
- Started writing dedicated documentation in the repository (https://github.com/mint-lang/mint/blob/time-rewamp/documentation/Core/Time.md)

All in all, it was fun working on this and figuring out the API and reading up on time-zones and stuff.

~This change will come in a later release before or with the syntax changes PR -  I want to release what's new in `master` as the next release.~

Read up on the new API in the documentation https://github.com/mint-lang/mint/blob/time-rewamp/documentation/Core/Time.md

------

Breaking changes:

- `Time` is now in UTC
- `Time.fromIso` => `Time.parseISO`
- `Time.toIso` => `Time.formatISO`
- `Time.toUnixTimestampInMs` => `Time.toUnix`
- `Time.fromUnixTimestampInMs` => `Time.unix`
- `Time.from` => `Time.utcDate`
- `Time.day` => `Time.dayOfMonth`
- `Time.relative` => `Time.distanceOfTimeInWords`
- `Time.format` - pattern changed from `YYYY` to percent ones `%Y` also now takes a language object
- Removed `Time.startOf`
- Removed `Time.endOf`

-----

~Not `Time` related changes (these can be ported over to `master` before merging this PR):~

- ~Use `make`s feature to only build the binary if actually dependent files are changed.~
- ~Added `format-core` task to format core files.~
- ~Added `Math.trunc` for integer division~
- Added `Number.isLeapYear` function (not sure if should be ported over)
- ~Added `String.padLeft` and `String.takeRight`~

-----

Any feedback is welcome:
- on naming conventions
- on new functions
- on missing functions
- on implementation details (happy to answer)

(@frnco I've would have added you as reviewer since you had interest in this topic, but can't because you're not in the organization)

### DIFF
--- a/core/source/Number.mint
+++ b/core/source/Number.mint
@@ -105,4 +105,8 @@ module Number {
       }
     }
   }
+
+  fun isLeapYear (year : Number) : Bool {
+    (year % 4 == 0) && (year % 100 != 0 || year % 400 == 0)
+  }
 }
--- a/core/source/Number.mint
+++ b/core/source/Number.mint
@@ -106,6 +106,12 @@ module Number {
     }
   }
 
+  /*
+  Returns if the given number (year) is a leap year or not.
+
+    Number.isLeapYear(2020) == true
+    Number.isLeapYear(2021) == false
+  */
   fun isLeapYear (year : Number) : Bool {
     (year % 4 == 0) && (year % 100 != 0 || year % 400 == 0)
   }
--- a/core/source/Number.mint
+++ b/core/source/Number.mint
@@ -105,14 +105,4 @@ module Number {
       }
     }
   }
-
-  /*
-  Returns if the given number (year) is a leap year or not.
-
-    Number.isLeapYear(2020) == true
-    Number.isLeapYear(2021) == false
-  */
-  fun isLeapYear (year : Number) : Bool {
-    (year % 4 == 0) && (year % 100 != 0 || year % 400 == 0)
-  }
 }
--- a/core/source/Time.mint
+++ b/core/source/Time.mint
@@ -1,31 +1,31 @@
-/* Represents a duration (span) of time. */
-enum Time.Span {
-  Milliseconds(Number)
-  Seconds(Number)
-  Minutes(Number)
-  Hours(Number)
-  Days(Number)
-  Weeks(Number)
-  Months(Number)
-  Years(Number)
-}
-
 /*
-Utility functions for working with `Time`.
+`Time` represents a date-time instant in incremental time observed in a specific
+time zone.
+
+The calendaric calculations are based on the rules of the proleptic Gregorian
+calendar as specified in ISO 8601. Leap seconds are ignored.
 
 This module uses the `Date`[1] JavaScript object under the hood. Since the
 `Date` object is always in the clients time-zone, this module modifies it's
 UTC version with the `getUTC*` and `setUTC*` methods.
 
 [1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date
+
+Implementation details:
+
+- Weekdays start from 1 (1 is Monday, 7 is sunday).
+- Months start from 1 (January).
+- Days start from 1.
 */
 module Time {
+  /* PARSING ----------------------------------------------------------------- */
+
   /*
   Tries to parse the given string as an ISO date.
 
-    Time.fromIso("2018-04-05T00:00:00.000Z")
+    Time.parseISO("2018-04-05T00:00:00.000Z")
   */
-  fun fromIso (raw : String) : Maybe(Time) {
+  fun parseISO (raw : String) : Maybe(Time) {
     `
     (() => {
       try {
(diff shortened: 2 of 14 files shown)


core/source/String.mint
```diff
@@ -311,4 +311,16 @@ module String {
...
+  fun padLeft (
+    padString : String,
+    targetLength : Number,
+    string : String
+  ) : String {
+    `#{string}.padStart(#{targetLength}, #{padString})`
+  }
+
+  fun takeRight (length : Number, string : String) : String {
+    `#{string}.slice(#{string}.length - #{length},#{string}.length)`
```
Sija, line 324:
```suggestion
    `#{string}.slice(#{string}.length - #{length}, #{string}.length)`
```

core/source/Time.mint
```diff
@@ -21,185 +36,617 @@ module Time {
...
+    Time.utc(2018, 4, 5)
+  */
+  fun utc (
+    year : Number,
+    month : Number,
+    day : Number,
+    hour : Number,
+    minute : Number,
+    second : Number,
+    millisecond : Number
```
Sija, line 73: All of these should be optional.

core/source/Time.mint
```diff
@@ -21,185 +36,617 @@ module Time {
...
   */
-  fun day (date : Time) : Number {
-    `#{date}.getUTCDate()`
+  fun year (time : Time) : Number {
+    `#{time}.getUTCFullYear()`
   }
 
   /*
-  Returns the UTC month of the given time.
+  Returns the quarterOfYear of the year in which the given time occurs.
```
Sija, line 149:
```suggestion
  Returns the quarter of the year in which the given time occurs.
```

core/source/Time.mint
```diff
@@ -21,185 +36,617 @@ module Time {
...
-          return DateFNS.endOfMonth(#{date})
-        case 'week':
-          return DateFNS.endOfWeek(#{date}, { weekStartsOn: 1 })
-        case 'day':
-          return DateFNS.endOfDay(#{date})
-        default:
-          return #{date}
-      }
+      const _ = #{time}.getUTCDay()
+      return _ === 0 ? 7 : _;
```
Sija, line 264: Please, avoid using `_` in JS-land as it may potentially interfere with utility libraries (and it's unreadable).

```suggestion
      const dayNumber = #{time}.getUTCDay()
      return dayNumber === 0 ? 7 : dayNumber;
```

core/source/Time.mint
```diff
@@ -21,185 +36,617 @@ module Time {
...
+  }
+
+  /*
+  Returns if the year of the given time is a leap year or not.
+
+    Time.isLeapYear(Time.utcDate(2011,1,1)) == false
+    Time.isLeapYear(Time.utcDate(2012,1,1)) == true
+  */
+  fun isLeapYear (time : Time) : Bool {
+    Number.isLeapYear(year(time))
```
Sija, line 357:
```suggestion
    time |> year |> Number.isLeapYear
```

core/source/Time.mint
```diff
@@ -21,185 +36,617 @@ module Time {
...
+  }
+
+  /* MANIPULATION ------------------------------------------------------------ */
+
+  /*
+  Shifts the given time using the given time span.
+
+    Time.shift(Time.utcDate(2018, 4, 5), Time.Span::Days(2)) ==
```
Sija, line 365:
```suggestion
    (Time.utcDate(2018, 4, 5) |> Time.shift(Time.Span::Days(2))) ==
```

core/source/Time.mint
```diff
@@ -21,185 +36,617 @@ module Time {
...
-    })
+    (() => {
+      const time = new Date(+#{time});
+
+      #{
+        case (delta) {
+          Time.Span::Milliseconds(amount) =>
+            `time.setUTCMilliseconds(time.getUTCMilliseconds() + #{amount})`
+
+          Time.Span::Seconds(amount)      =>
```
Sija, line 378:
```suggestion
          Time.Span::Seconds(amount) =>
```

ditto elsewhere

core/source/Time.mint
gdotdesign: I agree, but Mint doesn't have optional arguments (yet).

core/source/Time.mint
gdotdesign: I think it's more readable my way, this seems to me an overuse of the pipe.

Sija: I'd assume that any expression with more than 1 call in the chain should be using pipes for expressiveness, but maybe that's just me 🤷‍♂️

core/source/Time.mint
Sija: Doc example should be updated then.

gdotdesign: Where do you see that in the docs? There is `Time.utc` with all parameters and `Time.utcDate` with just year, month, day.

Sija: `Time.utc(2018, 4, 5)`

core/source/Time.mint
ryanprior: Is there a performance cost to the pipe, or is it just syntax?
Could we get an arrow that works like Clojure's, without needing a pipe between each step? This would look beautiful as
```
time -> year Time.isLeapYear
```

(Also with regard to the above: why not put `isLeapYear` in the `Time` module? It doesn't seem to have anything to do with Numbers generally.)

core/source/Time.mint
ryanprior: Both these look equally nice to me, but neither looks near as nice as Crystal's `+ 2.days`, wish we could capture a greater economy of expression here.
Since `Time.shift` can only accept a `Time.Span::T(n)`, could we perhaps allow the compiler to accept a plain `T(n)` which is automatically cast as a `Time.Span`? That would at least give us
```
Time.utcDate(2018, 4, 5) |> Time.shift(Days(2))
```

Then picture another pipe syntax `|.` which automatically keeps us in the same module (`Time` in this case) as the last expression. That gets us down to:

```
Time.utcDate(2018, 4, 5) |. shift(Days(2))
```

Then we're getting somewhere.

frnco: > Both these look equally nice to me, but neither looks near as nice as Crystal's `+ 2.days`, wish we could capture a greater economy of expression here. Since `Time.shift` can only accept a `Time.Span::T(n)`, could we perhaps allow the compiler to accept a plain `T(n)` which is automatically cast as a `Time.Span`? That would at least give us
> 
> ```
> Time.utcDate(2018, 4, 5) |> Time.shift(Days(2))
> ```
> 
> Then picture another pipe syntax `|.` which automatically keeps us in the same module (`Time` in this case) as the last expression. That gets us down to:
> 
> ```
> Time.utcDate(2018, 4, 5) |. shift(Days(2))
> ```
> 
> Then we're getting somewhere.

`Days(2)` looks nice at first but it would require leaving a bunch of classes unscoped, i.e., `Hours`, `Minutes`, `Seconds`, `Days`, `Weeks`, `Months` and `Years`. Time.days(2) seems better to me, and internally it could make the conversion, i.e.:
```
module TIme {
  fun days(amount : Number) : Time.Span::Days {
    Time.Span::Days(amount)
  }
}
```

That would also solve the problem with Weeks, as `Time.weeks` could use days in the background, i.e.:
```
module TIme {
  fun weeks(amount : Number) : Time.Span::Days {
    Time.Span::Days(amount * 7)
  }
}
```

frnco:
One consideration: The internationalization is done by using Enums and functions in the constant `ENGLISH`. I'd consider changing that to simpler `Number`s, maybe using an Array or something, i.e.:
```
module Time {
  const ENGLISH_MONTHS = ["January","February","March","April","May","June","July","August',"September","October","November","December"]
  fun monthName(month : Number) : String {
    case(maybeMonthName) {
      Maybe::Nothing => "January" /* Should never happen but we need this */
      Maybe::just(someMonthName) => someMonthName
    }
  } where {
    maybeMonthName
    (month - 1)
    |> Math.clamp(0,11)
    |> Array.at(ENGLISH_MONTHS)
  }
}
```

As it stands, to write translations one would need to duplicate a lot of code, including `case` branches, which seems kinda wasteful.

A better alternative, albeit dependent on adding yet another new data type, would be:
```
record Time.Names.Months {
  january : String,
  february : String,
  march : String,
  april : String,
  may : String,
  june : String,
  july : String,
  august : String,
  september : String,
  october : String,
  november : String,
  december : String
}
module Time {
  const ENGLISH_MONTHS = Time.Names.Months("January","February","March","April","May","June","July","August',"September","October","November","December")
  fun monthName(names : Time.Names.Months, month : Month) : String {
    case(month) {
      Month::January => names.january
      Month::February => names.february
      Month::March => names.march
      Month::April => names.april
      Month::May => names.may
      Month::June => names.june
      Month::July => names.july
      Month::August => names.august
      Month::September => names.september
      Month::October => names.october
      Month::November => names.november
      Month::December => names.december
    }
  }
}
```
But that would mean calling the fuction using `Time.monthName(Time.ENGLISH_MONTHS, SomeMonth)`, which also has its problems.

Still, I'd try to go for something that, like the examples I've given, would reduce the necessary boilerplate when defining translations.

core/source/Time.mint
ryanprior: It would still end up as `|> Time.shift(Time.days(2))` - repeating `Time` over and over. I have this pattern frequently in my code where I repeat the same module name ad nauseum, so I think it would be appropriate to add a language-level feature for brevity.

frnco: > It would still end up as `Time.utcDate(2018, 4, 5) |> Time.shift(Time.days(2))` - repeating `Time` over and over. I have this pattern frequently in my code where I repeat the same module name ad nauseum, so I think it would be appropriate to add a language-level feature for brevity.
> 
> Compare to Crystal's `Time.utc(2018, 4, 5) + 2.days` - no ambiguity there and no repetition. I'm always softly dismayed when I see something that lends itself to simple, clear representation in Crystal but it ends up boilerplatey and obfuscated in Mint.

It's hard to find examples that adequately portray what we're thinking, and we have to decide how much time we're willing to invest in picking/perfecting examples and when it's sufficient, but luckily I'm mostly a Ruby programmer so it's ridiculously easy to understand your point, which is all about the benefits of having a chaining/object-oriented approach.

In Rails I can simply write `Time.utc(2018,4,3) + 2.days` and it'll be equal to April 5th, 2018, UTC. Mint favors the functional approach, i.e., return values should be passed to functions, not be used to call functions using `.something`. There are pros and cons for both of those approaches, but it is what it is, Crystal is a good example of the absolute opposite approach, I've actually participated in a discussion about implementing pipes in Crystal (Like the `|>` operator in Mint) and there's a lot of resistance against it, because Crystal is all about OO and method chaining. A spin-off of Mint that had those would be something VERY interesting, but I, as many, currently lean towards  the Functional approach, which I like quite a lot, but it does, indeed, have a few issues. And Boilerplate is indeed one of the biggest problems, I believe. I avoided going into more detail to keep it short and focus on the `Time` module which is the topic here, but this is so important it deserves some clarification, so let's get to it:

Scoping is very important, this isn't obvious when talking about `Days` but if we consider a slightly larger system, let's take a simple blog/website as an example, it gets icky. Because imagine you start with `Post`s and `Page`s, plus the basic stuff, then you add `Thumbnail`s and `Category`(ies) for `Post`s, and `User`s. Then you decide on having `Thumbnail`s for `User`s, the name is taken already but it's ok, you just use `Avatar`, but then your boss tells you they want to add thumbnails for the categories, and you realize there's no sane way of making it work without having any ambiguity and with everything being easily recognizable and traceable and obvious. If you used `Post.Thumbnail`, `User.Thumbnail` and `Post.Category` it's actually not a problem, you just add `Post.Category.Thumbnail` and it never even poses any problem at all. That's the magic of scoping stuff, and it's important, and I actually believe we shouldn't mess with that.

Still, there is a nice option that would make both of us happy with the result, which is allowing access to scoped stuff withoug specifying the whole scope. Elm does that magnificently, and I miss how easy Elm makes to write `match`/`case` statements. Translating to Mint it'd be as simples as writing:
```
case(maybeSomething) {
  Just(something) => doSomethingWith(something)
  Nothing => doSomethingElse()
}

```
Notice `Maybe::Nothing` turned into a much simple `Nothing`. In Elm, if you write it like that and it finds more than one definition of `Nothing` it'll complain at compile-time and tell you to write the scope properly. Returning to the Blog example I've laid out above, that'd mean you'd only need to write the full scope when calling a Thumbnail, and you'd simply write `Category` and let the compiler figure out it's full/scoped name is actually `Post.Category`. I think if we want to avoid repetition in Mint we should probably go this way, instead of considering un-scoping methods on a case-by-case-basis, which will not only greatly increase the chances of something bad happening, but also results in re-enacting this discussing for every case, which compounds into a lot of time. The new `|.` operator was not a bad suggestion, but I think it's still not as good as a way to access unscoped methods/classes/enums/etc., and actually there's already something like that in Mint, when connecting Stores, as seen in https://www.mint-lang.com/guide/reference/components/connecting-stores (i.e., the line `connect Counter exposing { count as myCount }`), it could be solved by something like `from Time unscope {seconds,minutes,hours,days,weeks,months,years}`, or like Rust does it, `use Time::{seconds,minutes,hours,days,weeks,months,years}`, which could be even easier if those were in a separate scope and we had the option of using `*` to match everything in that scope, which could turn it into something as short as `use Time::Span::{*}` or, using the first notation that occurred to me, `from Time::Span unscope {*}`. That's how I'd approach this, which would fix all the cases we've been talking about where the pipes stay in the same module, but beyond that it'd solve the situation where you wanna pipe stuff to another module, then some other module, then back etc., which wouldn't be improved at all with the `|.` operator. This is not to say it's not a good idea to have the proposed `|.` operator, maybe it's a good ideal to ALSO have it, I just think that, first, a way to unscope methods has more applications and deserves a higher priority, and second, my understanding of the impacts, uses, advantages and possible issues with having the `|.` operator is insufficient. I don't know what languages have something like that, how people use it, basically what are the implications of having something like that in a language. I do have experience with Elm and Rust which provide ways to omit the scopes, and I can say with confidence that it makes things a lot easier.

If you do have examples, experience, reasons, information or whatever that leads you to believe something like `|.` is still something good to have, I encourage you to open an issue about that, but in the event it was just some wild brainstormy idea about a way to make the code less repetitive, we'll probably agree it's safer to use other languages as a reference before deciding what is a better fit for Mint. Personally, I'd go for a variation of `connect Store exposing {methods}`, only instead of a Store it would apply to a Module (Or whatever else this may apply to, though I can't recall anything other than Modules).

Either way, I do think this deserves an issue, and if you don't open one regarding either the `|.` operator or a way to reduce repetition, I'll create one after polishing this idea a bit and checking how I might go about implementing it.

core/source/Time.mint
gdotdesign: > Is there a performance cost to the pipe, or is it just syntax?

It's just syntax.

> Could we get an arrow that works like Clojure's, without needing a pipe between each step? This would look beautiful as

I'm working on something that looks like this:

```
time.year().isLeapYear()
```

It's how Nim does it (https://nim-lang.org/docs/manual.html#procedures-method-call-syntax). It allows the use of OO like notation for functions defined in modules which is basically a pipe without the module name.

I'll create a discussion or PR soon for this.

core/source/Time.mint
gdotdesign: No need to have either a new operator `|.` or a new language feature for importing thinks in scope.

In this branch https://github.com/mint-lang/mint/tree/syntax-rewamp-3 there are features that tackle these issues:

- Short enum options `:Day(1)` - it will raise an error if it's ambiguous
- OO like function calls  `["a"].first().withDefault("") == "a"` - it will raise an error if it's ambiguous

Like I said in the other reply, soon I'll create a PR or discussion about these changes, but for now let's keep the discussion here about the `Time` module.

frnco:
Important point about translation, I've just checked it and:

- `core/source/Time/Languages/English.mint` has 185 lines.
- Of those, 72 have Strings (Anything surrounded by `"`quotes`"`)
- For a translator, this means to translate 72 strings they need to write 185 lines of actual code, or copy-paste the code and replace the strings.

In case anyone is curious about how much I know what I'm talking about, some background: My mother graduated in Journalism, my father worked in Newspapers (On of which is where he met my mother, BTW), my brother works as a writer, translator and text reviewer, one of his Jobs was in the biggest newspaper of the seconds biggest state in Brazil (For completeness sake, the newspaper is Estado de Minas, the state is Minas Gerais), I could mention uncles in many fields from teachers to translators to educators, and I myself worked in PR agencies and as a Freelance contractor, and actually did quite a bit of translation work myself. I was also the lead translator for [Concrete5 ](https://www.concretecms.com/) for some time, dir volunteer translations for fansubs and the list goes on.

My point is: Translation is HARD. And the flexibility of doing it in code is actually a good thing since some languages have very unique features. But having to deal with over twice the number of lines is definitely a bad thing, and ufortunately most people that could be hired as translators will have a VERY hard time working with this. If we're aiming to include as many translations as reasonably possible IN the core api, it could be a good idea to go this way, but if the aim is for people to write their own translations then this needs quite a bit of work.

I'd personally favor using records as much as possible (Months/Weekdays are the most obvious, it drops the lines in the translation file from 58 to 23, but it's probably a good idea to also have those for most stuff). Ideally, the logic (`seconds < 30`, `minutes < 2` etc) would be shared by all translations, i.e.:
```
          if (minutes < 1) {
            if (positive) {
              language.someSecondsAgo
            } else {
              language.inSomeSeconds
            }(seconds)
          }
```
would become something more like
```
          if (minutes < 1) {
            if (positive) {
              if (seconds < 30) {
                seconds
                |> Number.toString
                |> interpolateString(language.secondsAgo.few)
              } else {
                seconds
                |> Number.toString
                |> interpolateString(language.secondsAgo.many)
              }
            } else {
              if (seconds < 30) {
                seconds
                |> Number.toString
                |> interpolateString(language.inSeconds.few)
              } else {
                seconds
                |> Number.toString
                |> interpolateString(language.inSeconds.many)
              }
            }
          }
```
With the addition of a function:
```
fun interpolateString(stringTemplate : Tuple(String,String), value : String) : String {
  "#{before}#{value}#{after}"
} where {
  {before,after} = stringTemplate
}
```

Though I suppose I'd prefer having a `record` instead of a Tuple, but the idea is the same: The translator would just deal with records/tuples actually related to the translation itself, without having to deal with any of the logic involved. This DOES pose a problem, especially with bigger numbers, as languages have their peculiarities, and trying something like this on the `Number` module would be a challenge. To mention a pretty interesting case that shows how this might pose a problem, in Japanese they have words for ten-thousands (man / まん) and hundred-millions (oku / おく), to mention a couple (Interesting read BTW: https://en.wikipedia.org/wiki/Japanese_numerals).

Even so, translation is already hard enough as it is, and asking translators to deal with code on top of that feels pretty bad. I have cloned this PR and am working on a simpler structure that may help with this, plus some other simplifications. @gdotdesign mentioned https://github.com/mint-lang/mint/tree/syntax-rewamp-3 having some syntactic sugar that could also help make it easier to use this so I'm also taking a look at that. I quite like having only `Time.shift` and using negative numbers, building my plans with the premises of having `before`/`after` made me blind to how easier it is to work like this. I think there are a lot of things the we can improve here, but this is some amazing improvement right here.

I'll share some ideas as soon as possible and hopefully we can make this even more polished.

gdotdesign:
> My point is: Translation is HARD. And the flexibility of doing it in code is actually a good thing since some languages have very unique features. But having to deal with over twice the number of lines is definitely a bad thing, and ufortunately most people that could be hired as translators will have a VERY hard time working with this. If we're aiming to include as many translations as reasonably possible IN the core api, it could be a good idea to go this way, but if the aim is for people to write their own translations then this needs quite a bit of work.

I don't think translators will write these translations, it will probably be programmers that need that specific language for their project. I see where you are getting at though, another way would be to use a JSON or YAML `@inline` it and then decode it to a record.

Furthermore, I have added the constant for the Hungarian language (since it's mother tongue), and it was good that it was in code, since I could write `toOrdinalSuffix` using a case since it differs from number to number.

>  I have cloned this PR and am working on a simpler structure that may help with this, plus some other simplifications.

@frnco did you make any progress? :)

------------------

@sija I would like to merge this PR soon, would you like me to separate the not time related bits into a different PR?

core/source/Time/Languages/Hungarian.mint
```diff
@@ -0,0 +1,206 @@
+module Time.Format {
+  /* The English language record for time formatting. */
+  const HUNGARUAN =
```
Sija, line 3:
```suggestion
  const HUNGARIAN =
```

Sija:
> @Sija I would like to merge this PR soon, would you like me to separate the not time related bits into a different PR?

Yes please, that would be great!

core/source/Number.mint
```diff
@@ -105,4 +105,14 @@ module Number {
...
+
+  /*
+  Returns if the given number (year) is a leap year or not.
+
+    Number.isLeapYear(2020) == true
+    Number.isLeapYear(2021) == false
+  */
+  fun isLeapYear (year : Number) : Bool {
+    (year % 4 == 0) && (year % 100 != 0 || year % 400 == 0)
+  }
```
Sija, line 117: This method IMO conceptually belongs more to the `Time` module.

gdotdesign: Something like `isNumberLeapYear`? Because `Time` already has a `isLeapYear` function and Mint doesn't have function overloading, that's why I moved it to the `Number` module.

Sija: Yup, sounds about right.

core/source/Time.mint
```diff
@@ -21,185 +36,627 @@ module Time {
...
+  */
+  fun isLeapYear (time : Time) : Bool {
+    isNumberLeapYear(year(time))
+  }
+
+  /*
+  Returns if the given number (year) is a leap year or not.
+
+    Number.isNumberLeapYear(2020) == true
+    Number.isNumberLeapYear(2021) == false
```
Sija, line 364:
```suggestion
    Time.isNumberLeapYear(2020) == true
    Time.isNumberLeapYear(2021) == false
```

Sija (approved):
🕚 to 🚢

Namek:
@gdotdesign is there a replacement for `Time.month`?

gdotdesign:
> @gdotdesign is there a replacement for `Time.month`?

`Time.monthNumber(time)` (1 based) is the replacement.

PR #528, opened 2021-12-16
More Syntax Changes

gdotdesign:
This is the second set of big syntax changes, it's a separate PR, so we can do this in parts, and it's pointing to the `code-blocks` branch for readability purposes since it'b based on that. 

### Removed Partial Application 

Since Mint isn't a typical functional language it doesn't really make sense to have partial application without other functional language features like operators for function composition. It also blocks the other two features introduced in this PR.

### Default Arguments 

You can now add default arguments to functions:

```
fun greet (name : String, shouldUppercase : Bool = false) {
  result = 
    "Hello #{name}!"

  if (shouldUppercase) {
    String.toUppercase(result)
  } else {
    result
  }
} 
```

- Optional arguments must appear at the end of the argument list after any non default arguments
- Optional arguments can be defined for **functions**, **inline functions** and **styles**

### Short Enum Options

It is now possible to use enum options without specifying the enum itself:

```
case (Just(value)) {
  Just(a) => "Got something!"
  Nonthing => "Damn!"
}
```

- Having this can lead to ambiguous code but makes the code more readable
- ~We need the `:` prefix because otherwise it would conflict with record constructors which I think we could remove from the language, but I need feedback to decide that~

### OO Style Calls

It is now possible to call functions in an object-oriented fashion.

```
module String {
  fun greet (name : String) {
    "Hello #{name}!"
  }
}

"Joe".greet()
```

This works like this:

1. We take all functions and select the ones that match the name
2. We select those functions where the last argument matches the type of the accessed entity (left-hand size in this case `"Joe"`)
3. If there are none or more than one we show an error
4. If there is one we can call that with the left-hand side as the last argument (like using the pipe operator)

There are benefits to this:

- Makes code super readable
- Reduces boilerplate
- Code looks and feels more like object-oriented code 

There are drawbacks to this as well:

- It can introduce ambiguity since there can be many functions with the same name and same last argument
- It introduces a slight overhead when compiling (since we need to check all functions)
- It mixes record access with calls which can be misleading (it's a function on the record or function in a module?)

----------

PS: There might be other changes in this PR in the future.


bellebethcooper:
I love these changes! Especially the short enum options. I use enums like this in Swift all the time, and I definitely prefer it.

gdotdesign:
I'm closing this because parts of it were cherry-picked into #503 and some will be not implemented at this point (OO style calls).

PR #530, opened 2022-01-10
Changes before the new `Time` module.

gdotdesign:


core/tests/tests/String.mint
```diff
@@ -222,3 +222,15 @@ suite "String.indentWithOptions" {
...
 }
+
+suite "String.padLeft" {
+  test "it pads the string from the left" {
+    String.padLeft("0", 2, "5") == "05"
+  }
+}
+
+suite "String.takeRight" {
+  test "it pads the string from the left" {
```
Sija, line 233: Copy-pasted spec name.

Sija (approved):
LGTM

core/tests/tests/Math.mint
```diff
@@ -80,6 +80,12 @@ suite "Math.fmod" {
   }
 }
 
+suite "Math.truncate" {
+  test "it truncates number to given amount" {
```
Sija, line 84: Copy pasted from `Math.truncate`.

gdotdesign: I don't understand your comment :) please elaborate.

Sija: It's copy pasted from `Math.truncate` specs - which is a different method.

gdotdesign: Ah, I see now the name should be `Math.trunc`.

PR #531, opened 2022-01-13
[LS] Code action framework and order entities code action.

gdotdesign:
This pull request adds the framework for handling [code actions] in the language server and a code action to reorder the entities (constants and functions) in a module. I added this first because I'm going through the core library and didn't want to manually reorder them :joy:

### DIFF
--- a/src/ast/node.cr
+++ b/src/ast/node.cr
@@ -40,8 +40,8 @@ module Mint
         end
       end
 
+      property from : Int32
       getter input : Data
-      getter from : Int32
       getter to : Int32
 
       def initialize(@input, @from, @to)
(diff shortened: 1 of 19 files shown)


src/ast/node.cr
```diff
@@ -40,8 +40,8 @@ module Mint
         end
       end
 
+      property from : Int32
```
gdotdesign, line 43: This needed to change because it's used for the reordering.

Sija (approved):
LGTM

src/ast/node.cr
Sija: I'm not following why this needs to be declared earlier?

gdotdesign: Not earlier, it changed from a `getter` to a `property`.

gdotdesign: It's earlier because of my slight OCD of having longer lines at the top :joy:

PR #537, opened 2022-02-24
Parser optimizations

asterite:
Hi!

See discussion starting at https://github.com/mint-lang/mint/pull/536#issuecomment-1049869835

This is just a draft... I think the style or final code can be further improved, but at least the optimization ideas should be clear from the commits.

This is the benchmark I used:

```crystal
require "./src/all"
require "benchmark"

filename = "icons.mint"
source = File.read(filename)

Benchmark.ips do |x|
  x.report("parse") do
    Mint::Parser.parse(source, filename)
  end
end
```

The file "icons.mint" comes from here: https://github.com/mint-lang/mint-tabler-icons/blob/as-functions/source/Icons.mint

Before this PR:

```
parse 108.63m (  9.21s ) (± 0.00%)  7.33GB/op  fastest
```

With this PR:

```
parse  68.05  ( 14.70ms) (± 3.80%)  20.1MB/op  fastest
```


gdotdesign:
Thank you very much :pray: I totally missed that `raise`, we have run into this before :facepalm: I'll take a look at other optimizations you made to learn from it :)

asterite:
Also with the [previous benchmark](https://github.com/mint-lang/mint/pull/536):

```crystal
require "./src/all"
require "benchmark"

filename = "./core/source/Provider/ElementSize.mint"
source = File.read(filename)
Benchmark.ips do |x|
  x.report("parse") do
    Mint::Parser.parse(source, filename)
  end
end
```

Before:

```
parse   8.55k (117.00µs) (± 4.06%)  83.9kB/op  fastest
```

After:

```
parse  15.92k ( 62.82µs) (± 6.46%)  37.2kB/op  fastest
```

So it's about twice as fast and consumes half the memory in the general case.

And for this particular `icons.mint` file, the speedup i like 600x faster and like 360 less memory consumed.

I still have to investigate why it takes 20MB to parse that file, so there's still room for optimization! But it can be done in a separate PR.

src/parser.cr
```diff
@@ -45,12 +46,8 @@ module Mint
     # ----------------------------------------------------------------------------
 
     def raise(error : SyntaxError.class, position : Int32, raw : Hash(String, T)) forall T
-      to =
-        input[position, input.size]
-          .join
-          .split(/\s|\n|\r/)
-          .first
-          .size
```
asterite, line 53: Imagining `raise` being raised on different places while parsing the code. This code will essentially take the rest of the chars, join them, then split them by whitespace and only take the first one. With a file of, say, 500K chars, if an error happens very early (an error for backtracking, like what happened when parsing parentheses) this would mean generating a 500K string, splitting it (generating an array with this huge string split into smaller strings)

So I changed it to scan for the next whitespace, then computing the distance. No memory allocations at all! :-)

src/parser.cr
```diff
@@ -59,7 +56,7 @@ module Mint
           input: data)
 
       part =
-        input[position, to].join
+        @string[position, to]
```
asterite, line 59: If we store the original string we can compute substrings by calling `String#[]`, which is much faster than getting a subrange of `Array(Char)` and then joining that.

Maybe it's worth introducing a helper method for this? I don't see instance variables being used anywhere in the parser.

src/parser.cr
```diff
@@ -93,6 +90,14 @@ module Mint
     # Consuming characters
     # ----------------------------------------------------------------------------
 
+    def letters_or_numbers
+      chars { |char| char.ascii_letter? || char.ascii_number? }
+    end
+
+    def letters_or_numbers_or_dash
+      chars { |char| char.ascii_letter? || char.ascii_number? || char == '-' }
+    end
```
asterite, line 99: I added these because they are used more than once. Feel free to suggest other names! Or we can inline that logic in the caller. But maybe it's good to have these named methods, it might also make the code slightly clearer.

src/parser.cr
```diff
@@ -104,18 +109,30 @@ module Mint
...
 
-    def chars(set) : String?
-      consume_while char != '\0' && char.in_set?(set)
+    def chars(& : Char -> Bool)
+      while char != '\0' && (yield char)
+        step
+      end
+    end
+
+    def chars_except(& : Char -> Bool)
```
asterite, line 128: Maybe this should be called `chars_until`?

src/parsers/basic_expression.cr
```diff
@@ -33,7 +33,8 @@ module Mint
         parallel ||
         try_expression ||
         case_expression ||
-        parenthesized_expression_or_inline_function ||
+        parenthesized_expression_no_error ||
```
asterite, line 36: It seems that the error produced by `parenthesized_expression` is only checked in the specs. But the actual parser just swallowed those errors and tried to parse an inline function. So now there's an unused `parenthesized_expression` (well, still used in specs) so I'm not sure this is an ideal state. Let me know!

src/parsers/js.cr
```diff
@@ -30,17 +30,15 @@ module Mint
     end
 
     def not_interpolation_part(terminator : Char, stop_on_interpolation : Bool = true) : String?
-      # What characters to match
-      chars_to_match =
+      value =
         if stop_on_interpolation
-          "^#{terminator}#" # Until we find either a terminator or interpolation
```
asterite, line 36: Doing this interpolation on every `not_interpolation_part` was very slow! Now there's no interpolation anymore :-)

src/parsers/variable.cr
```diff
@@ -3,8 +3,8 @@ module Mint
     def variable_attribute_name : Ast::Variable?
       start do |start_position|
         value = gather do
-          next unless char.in_set? "a-z"
-          chars "a-zA-Z0-9:-"
+          next unless char.ascii_lowercase?
+          chars { |char| char.ascii_letter? || char.ascii_number? || char == '-' || char == ':' }
```
asterite, line 7: I didn't extract this one to a method because it's the only occurence

src/parsers/basic_expression.cr
gdotdesign: I guess it changed at some point, so If it's not used in the parser itself, we can remove it and have only one `parenthesized_expression` which returns ˙Ast::ParenthesizedExpression?`

src/parser.cr
gdotdesign: I like having a helper method for this, not sure what to call it.

src/parser.cr
asterite: Done!

src/parser.cr
asterite: I added one called `substring`

src/parsers/basic_expression.cr
asterite: Done! This also removed some error types that apparently were unused.

src/parser.cr
```diff
@@ -7,7 +7,8 @@ module Mint
     getter refs = [] of {Ast::Variable, Ast::HtmlComponent | Ast::HtmlElement}
     getter position = 0
 
-    def initialize(input, @file)
+    def initialize(input : String, @file)
+      @string = input
```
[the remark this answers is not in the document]
asterite: Done!

gdotdesign (approved):

src/parser.cr
```diff
@@ -233,5 +261,24 @@ module Mint
 
       result
     end
+
+    # Getting substrings out of the original string
+    # ----------------------------------------------------------------------------
+
+    def substring(from, to)
```
Sija, line 268:
```suggestion
    # Gets substring out of the original string
    def substring(from, to)
```

asterite: Oh, I see. I was just following this style of comments that are in this file. I'll change it!

src/parser.cr
```diff
@@ -233,5 +261,24 @@ module Mint
...
+      @data.input[from, to]
+    end
+
+    private def next_whitespace_index
+      whitespace_index = nil
```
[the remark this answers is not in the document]
asterite: Oh, you are right. I think I did some refactors and it ended up different and wrong. I also guess there are no specs for this 🤔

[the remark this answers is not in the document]
Sija: Since all of the specs passed with this method returning `nil` at all times, would you mind adding a spec for this? 😅

Sija (approved):
LGTM 🚀 (missing spec for `next_whitespace_index` would be a nice to have, but not required)

src/parser.cr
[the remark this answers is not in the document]
gdotdesign: I think writing a simple test which parses something that raises and produces an error, and then checking the error if the passed value is what we want.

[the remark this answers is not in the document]
asterite: Would it be okay if any if you writes that spec? I _could_ do it, but I already spent some time for this today and I have other things to do. I feel like one of you could write that spec much faster than me.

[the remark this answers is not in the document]
gdotdesign: Yep, no worries, I'll get to it some time :sweat_smile:

gdotdesign:
Thank you very much! :pray: :tada:

PR #539, opened 2022-03-15
Fix service-worker "fetching" issue 

Eternahl:
Issue #370


Eternahl:
I don't think there's anything I can do about the failure of this "checks"
It looks like it's just an authentication issue

```
Downloading Crystal build from https://dl.circleci.com/private/output/job/6ff8e112-da7b-4833-ba05-38428295f826/artifacts/0/dist_packages/crystal-1.4.0-dev-1-darwin-universal.tar.gz
Error: Error: Unexpected HTTP response: 401
```

src/builder.cr
```diff
@@ -174,8 +174,24 @@ module Mint
...
+        Assets.read("sw-utils.js")
+      end
+    end
+
     def terminal
       Render::Terminal::STDOUT
     end
+
+    def service_worker(artifacts, relative, optimize, sw_utils_path)
+      worker = ServiceWorker.generate(artifacts, relative, optimize)
```
Sija, line 193:
```suggestion
      worker = ServiceWorker.new(artifacts, relative, optimize)
```

src/utils/service_worker.cr
```diff
@@ -1,12 +1,16 @@
...
   class ServiceWorker
     @relative : Bool
+    @optimize : Bool
+    @artifacts : TypeChecker::Artifacts
 
-    def self.generate(relative = false)
-      new(relative).to_s
+    def self.generate(artifacts, relative, optimize)
+      new(artifacts, relative, optimize)
     end
```
Sija, line 10: There's no point in keeping this method after these changes.

```suggestion
```

src/utils/service_worker.ecr
```diff
@@ -45,12 +46,21 @@ self.addEventListener('fetch', event => {
 
   // If we are on the same origin
   if (isSameOrigin) {
-    // Try to get the response from the cache, and if not available
-    // fall back to the "index.html" file.
+    // Try to get the response from the cache,
+    // if not available, check the routing,
+    //   if the route exist, serve "index.html" file
+    //   else fetch request
```
Sija, line 52:
```suggestion
    // if not available, check the routing, and if the route exist,
    // serve "index.html" file or fetch request otherwise
```

src/builder.cr
```diff
@@ -174,8 +174,24 @@ module Mint
...
+      end
+    end
+
     def terminal
       Render::Terminal::STDOUT
     end
+
+    def service_worker(artifacts, relative, optimize, sw_utils_path)
+      worker = ServiceWorker.new(artifacts, relative, optimize)
+      "#{worker}#{get_service_worker_utils(sw_utils_path)}"
```
Sija, line 194: Service worker utils should come first, I guess?

```suggestion
      "#{get_service_worker_utils(sw_utils_path)}#{worker}"
```

src/utils/service_worker.cr
```diff
@@ -37,5 +37,16 @@ module Mint
...
+
+    def get_routes : String
+      routes = Mint::Compiler
+        .new(TypeChecker.check(@artifacts.ast))
+        .compile_service_worker(@artifacts.ast.routes)
+        .map do |node|
+          "...#{node}"
+        end
+
+      @js.const("routes", "[#{routes.join(',')}]")
```
Sija, line 49:
```suggestion
      @js.const("routes", "[#{routes.join(", ")}]")
```

src/builder.cr
```diff
@@ -174,8 +174,24 @@ module Mint
       {runtime + compiled.to_s, type_checker.artifacts}
     end
 
+    def get_service_worker_utils(sw_utils_path)
```
gdotdesign, line 177: Won't `sw_utils_path` be the same as the `runtime_path`?

src/assets/sw-utils.js
```diff
@@ -0,0 +1,946 @@
+var Mint = (function (exports) {
```
Sija, line 1: Is defining `Mint` here is correct? Wouldn't that clash with the rest of the runtime?

gdotdesign: It won't because they are not loaded simultaneously, the `sw-utils.js` is loaded in the service worker while the `runtime.js` is loaded in the application itself.

src/builder.cr
Eternahl: In _Javascript_ the `var` declarations are hoisted (this is the case of the worker code), so it is not necessary.
But I don't mind the change.

src/assets/sw-utils.js
Eternahl: The _"standard runtime"_ is not included in the `service-worker` context, so it should not conflict with it.

src/builder.cr
gdotdesign: This is not an issue since the `Mint` variable is used later on, and it's in the global scope so it will be found. I do like that it's before though.

src/builder.cr
gdotdesign: Just tested this and the user provided runtime will be added to the service worker, but I'm not sure if it's an issue or not, the user provided runtime might have the `isMatchingRoute` function, but the custom runtime feature is not documented so :man_shrugging: 

It _feels like having a separate flag_ for the custom service worker runtime should be the way to go, but it's not necessary to implement it in this PR so you can remove that part.

gdotdesign:
I tested it using your branch, and it works well :+1: Apart from the custom runtime issue LGTM :pray:

src/builder.cr
```diff
@@ -85,7 +85,7 @@ module Mint
       unless skip_service_worker
         terminal.measure "#{COG} Creating service worker..." do
           File.write Path[DIST_DIR, "service-worker.js"],
-            ServiceWorker.generate(relative)
+            service_worker(artifacts, relative, optimize, runtime_path)
```
Sija, line 88:
```suggestion
            service_worker(artifacts, relative, optimize)
```

src/builder.cr
```diff
@@ -174,8 +174,24 @@ module Mint
...
+    def get_service_worker_utils(sw_utils_path)
+      if sw_utils_path
+        raise RuntimeFileNotFound, {
+          "path" => sw_utils_path,
+        } unless File.exists?(sw_utils_path)
+        File.read(sw_utils_path)
+      else
+        Assets.read("sw-utils.js")
+      end
+    end
```
Sija, line 186:
```suggestion
    def get_service_worker_utils
      Assets.read("sw-utils.js")
    end
```

src/builder.cr
```diff
@@ -174,8 +174,24 @@ module Mint
...
+
     def terminal
       Render::Terminal::STDOUT
     end
+
+    def service_worker(artifacts, relative, optimize, sw_utils_path)
+      worker = ServiceWorker.new(artifacts, relative, optimize)
+      "#{get_service_worker_utils(sw_utils_path)}#{worker}"
+    end
```
Sija, line 195:
```suggestion
    def service_worker(artifacts, relative, optimize)
      worker = ServiceWorker.new(artifacts, relative, optimize)
      "#{get_service_worker_utils}#{worker}"
    end
```

Sija (changes requested):
As per https://github.com/mint-lang/mint/pull/539#discussion_r841654010

gdotdesign (approved):

gdotdesign:
Thank you @Eternahl :pray:

PR #546, opened 2022-05-15
Upgrade crystal version to 1.4.1

alex-lairan:
Small PR to keep Crystal dependency up-to-date.

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -1,5 +1,5 @@
 name: mint
-version: 0.16.0
+version: 0.16.1
 
 authors:
   - Szikszai Gusztáv
@@ -30,6 +30,6 @@ targets:
   mint:
     main: src/mint.cr
 
-crystal: ~> 1.3.0
+crystal: ~> 1.4.1
 
 license: BSD-3-Clause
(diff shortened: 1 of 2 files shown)


shard.yml
```diff
@@ -1,5 +1,5 @@
 name: mint
-version: 0.16.0
+version: 0.16.1
```
Sija, line 2: No need to bump version. Please revert this.

shard.yml
```diff
@@ -30,6 +30,6 @@ targets:
   mint:
     main: src/mint.cr
 
-crystal: ~> 1.3.0
+crystal: ~> 1.4.1
```
Sija, line 33:
```suggestion
crystal: ~> 1.3
```

Sija (changes requested):

shard.yml
ryanprior: If you don't want to update to crystal 1.4 maybe just close the PR? I don't get why you'd request this change.

Sija: There's no need to bump minimal supported Crystal version. It has nothing to do with the _upgrading_ crystal, as it states what versions are supported, and mint support 1.3+.

Sija:
I changed the crystal version in `shard.yml`. The entry in `.tool-versions` I'll bump once crystal 1.5.0 is out.
Thanks for the push :)

PR #551, opened 2022-10-20
Update dependencies

straight-shoota:
Almost all dependencies are outdated. This patch updates all of them to their latest releases.

This removes a compiler warning from old kemal version which is fixed in the newest release.

I'd suggest loosening the version restrictions a bit to allow minor release progressions as well. That should work fine with these shards. But for now I kept the strong version pinning.


Sija (approved):

gdotdesign:
Thanks :pray:

PR #573, opened 2023-04-07
Fix/ls eof

jansul:
This is a "low touch" fix to the issue I have been facing where the language server will 100% CPU when not closed correctly (i.e. when the client doesn't send the ["Exit" lifecycle message](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#exit))

The way I have been replicating this is to reload VSCode via "Command Pallette > Developer Reload Window" while the language server is running.

The reason it happens is because we are continuously trying to read a message from `STDIN` even when there is nothing left to read from it (EOF). When it's in this state `gets` is no longer blocking causing high CPU.

The fix I have chosen is to use [`read_line`](https://crystal-lang.org/api/1.7.3/IO.html#read_line(*args,**options):String-instance-method) rather than `gets` which raises an `EOFError` rather than returning `Nil` continuing the loop.

---

There is one that is confusing for me, there is this line in the server which is run on every loop:

```crystal
return exit(1) if @in.closed?
```

I'm not quite sure why this doesn't catch it. If an file returns EOF shouldn't it also be closed? Then again I don't really know much about files. Would be nice to see if others can replicate the issue to rule out it being something with my environment.

---

I've added a very basic test for the change. I could probably add more for the actual server itself if the code so far is OK. Still very much getting used to Crystal :)


spec/lsp/message_parser_spec.cr
```diff
@@ -0,0 +1,38 @@
...
+      io = IO::Memory.new
+      io.print "Invalid-Header: Invalid\r\n"
+      io.print "Content-Length: #{body.bytesize}\r\n"
+      io.print "\r\n"
+      io.print body
+      io.rewind
+
+      content = LSP::MessageParser.parse(io) do |content|
+        content
+      end
```
Sija, line 33:
```suggestion
      content = LSP::MessageParser.parse(io, &.itself)
```

Sija (commented):
Few minor suggestions, otherwise looks good! ❤️

spec/lsp/message_parser_spec.cr
jansul: TIL about `.itself`, thanks! :smile:

gdotdesign:
I ran into this issue of the server not exiting correctly, and I thought this was the fix :sweat_smile: 

```crystal
return exit(1) if @in.closed?
```

so I guess we can remove it at some point, or just keep it as a fallback :grinning: 

I'm glad that you could track it down :+1:

gdotdesign (approved):

Sija (approved):

PR #575, opened 2023-04-15
LSP: Add support for Go to Definition

jansul:
This PR adds support for the LSP feature [Go to Definition](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_definition). Only a few things are actually supported, namely finding the definition of  `HtmlAttribute`, `HtmlComponent` and `HtmlStyle` nodes as demonstrated below:

[Screencast from 2023-04-15 16-37-04.webm](https://user-images.githubusercontent.com/1927518/232234627-8d20f691-c57f-47cf-8f1b-44ae60ee3459.webm)

I've opened this as a draft, as I'm not sure what I've done is a very good approach or not, and also there are some unfinished bits:

* I need to check if the client supports [`LocationLink`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#locationLink)'s before returning them (via `textDocument.definition.linkSupport`) and otherwise just return a [`Location`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#location)'s instead (also looks like I should be returning an array of them as per the [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_definition)?)

The main parts I would like feedback on are below:

--- 

### Selecting the "name" part of a node 

The LSP requires that you return both `targetRange` and `targetSelectionRange` as part of a [`LocationLink`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#locationLink) as described below:

```js
/**
* The full target range of this link. If the target for example is a symbol
* then target range is the range enclosing this symbol not including
* leading/trailing whitespace but everything else like comments. This
* information is typically used to highlight the range in the editor.
*/
targetRange: Range;

/**
* The range that should be selected and revealed when this link is being
* followed, e.g the name of a function. Must be contained by the
* `targetRange`. See also `DocumentSymbol#range`
*/
targetSelectionRange: Range;
```

Currently I've implemented `targetSelectionRange` by dumping a bunch of `selection` methods onto the `Definition` class. These are used to figure out which part of each node is the "name" (which can be a bit complicated for some nodes)

These methods could probably be moved to the actual Ast nodes themselves a bit like we have `.static?` etc?

e.g.

```crystal 
class Ast::Node
    def name : Ast::Node
    end
    # or
    def name_location : Ast::Node::Location
    end
end
```

---

### Finding other nodes in the "stack"

For figuring out the "definition" of a node, I needed to be able to look at other nodes in the stack as provided by `server.nodes_at_cursor`.

As an example, the stack when hovering over a `HtmlStyle` node might look like this:

```
Mint::Ast::Variable
 ↳ Mint::Ast::HtmlStyle
  ↳ Mint::Ast::HtmlElement
   ↳ Mint::Ast::Statement
    ↳ Mint::Ast::Block
     ↳ Mint::Ast::Function
      ↳ Mint::Ast::Component
```

To find the corresponding `Style` node, I also need to know which `Component` the `HtmlStyle` is in.

To do this, I added a bunch of methods via macro that allow you to look along the stack (`next_<node_name>` for finding an immediate parent node,  and `any_<node_name>` for any ancestor node), returning `Nil` if they can't find anything

In code this looks something like this:

```crystal
return unless variable =
                next_variable stack
return unless next_html_style stack
return unless component =
                any_component stack
```

I'm not sure if this is a very good or efficient approach, and am open to suggestions for a better one!


spec/language_server/definition_spec.cr
```diff
@@ -0,0 +1,45 @@
...
+end
+
+Dir
+  .glob("./spec/language_server/definition/**/*")
+  .select! { |file| File.file?(file) }
+  .sort!
+  .each do |file|
+    it file do
+      with_workspace do |workspace|
+        manifest, *rest = File.read(file).split("-" * 80)
```
gdotdesign, line 14: Oh, this is neat, didn't know Crystal supports this :metal:

gdotdesign:
> * I need to check if the client supports [`LocationLink`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#locationLink)'s before returning them (via `textDocument.definition.linkSupport`) and otherwise just return a [`Location`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#location)'s instead (also looks like I should be returning an array of them as per the [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_definition)?)

- Why do we use `LocationLink` instead of just `Location`?
- The return type can be many things `Location | Location[] | LocationLink[] | null` so returning a single location or link is OK

-----

While I think the macro implementation is cool :sunglasses: it can be written with just two methods: 

```
def next_node(klass : N.class, stack : Array(Ast::Node)) : N | Nil forall N
  stack.shift?.try &.as?(N)
end

def any_node(klass : N.class, stack : Array(Ast::Node)) : N | Nil forall N
  while node = stack.shift?
    if node.is_a?(N)
      return node
    end
  end
end
```

And then used as:

```
next_node Ast::Variable, stack
next_node Ast::HtmlComponent, stack
any_node Ast::Component, stack
```

I'm not sure, but it might be refactored in a way, so we wouldn't need to create duplicates and have less memory footprint (maybe @Sija has some ideas? :smile: )

------

Otherwise I think it's OK :+1:

spec/language_server/definition_spec.cr
jansul: Will probably change the tests to match the [type_checking ones](https://github.com/mint-lang/mint/blob/master/spec/type_checking/html_element_not_in_component) rather than having a manifest at the top for consistency, so this will might have to go 😅 

Have been enjoying my time with Crystal so far tho!

jansul:
> * Why do we use `LocationLink` instead of just `Location`?

I haven't tried just using only `Location` yet, but I believe you get the cool underline when hovering (like a hyperlink) when using `LocationLink` so I believe it's worth supporting it.

Shouldn't be too hard to fall back to using `Location` if it's not supported by the client.

---

> While I think the macro implementation is cool 😎 it can be written with just two methods:

Doh, it's easy to try and be too fancy with macros 🙈 I'll make this change!

---

What were your opinions on adding additional methods to `Ast::Node`?  (for finding the "name" part of a node) Or should I just keep that seperate for now?

gdotdesign:
> What were your opinions on adding additional methods to `Ast::Node`? (for finding the "name" part of a node) Or should I just keep that seperate for now?

I would keep it separate for now, let's see it a common pattern emerges and then we can add it accordingly.

jansul:
I think this is ready for review now :+1: 

I've made the changes above, as well as a few others:

* The tests are now closer in format to the [type_checking ones](https://github.com/mint-lang/mint/blob/master/spec/type_checking/html_element_not_in_component). They allow multiple requests and responses, and potentially could be used to test the other LS features
* A few bug fixes, such as pointing to an incorrect location when a component had comments

I've tested it on the recently updated `mint-docs-viewer` repository and it seems to work reasonable well. Let me know if you find any other bugs. :smile: 

---

One of the changes is that It now ignores any components that are part of the "core" set, mostly because there is nowhere on disk where vscode etc can find them.

To be able to test this. I removed these lines from `spec_helper.cr`:

```diff
- module Mint
-   class Core
-     class_getter ast : Ast = Ast.new
-   end
- end
```

Let me know if you would like this back in. I could probably just add a minimal set of components to this for testing rather than all of the core files.

spec/language_server/definition/core_component
```diff
@@ -0,0 +1,50 @@
...
+  "result": null,
+  "id": 0
+}
+------------------------------------------------------------------------response
+{
+  "jsonrpc": "2.0",
+  "result": null,
+  "id": 1
+}
+------------------------------------------------------------------------response
```
gdotdesign, line 50: Please add a new line to the end of every file that is missing it :)

src/ls/definition.cr
```diff
@@ -0,0 +1,166 @@
+module Mint
+  module LS
+    class StackReader
```
gdotdesign, line 3: We can move this into its own file (I like it :+1:)

src/ls/definition.cr
```diff
@@ -0,0 +1,166 @@
...
+      property params : LSP::TextDocumentPositionParams
+
+      def execute(server) : LSP::LocationLink | LSP::Location | Nil
+        uri =
+          URI.parse(params.text_document.uri)
+
+        workspace =
+          Workspace[uri.path.to_s]
+
+        unless workspace.error
```
gdotdesign, line 33: Do we need to return something in case we have an error?

src/ls/definition.cr
```diff
@@ -0,0 +1,166 @@
...
+        #   global component MintComponent {
+        #                    ^^^^^^^^^^^^^
+
+        start_line, start_column = node.location.start
+
+        node.comment.try do |comment|
+          start_line, start_column = comment.location.end
+        end
+
+        # TODO: Change parser so component name is a node?
```
gdotdesign, line 76: You can change the parser if you can :)

jansul: Would certainly simplify this method :smiley: 

Looking at it, the best way might be to have the `type_id` methods return a new node (`TypeId`?) rather than a `String`?

src/ls/definition.cr
jansul: I think we should probably be returning a [`ResponseError`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#responseError) for these errors, but I think that would require a bigger refactoring of the `LSP::Server` :sweat_smile: 

Without that, I think `null` is probably the most appropriate out of the definition responses (i.e.  ```Location | Location[] | LocationLink[] | null```)

src/ls/definition.cr
gdotdesign: Oh I forgot that it was a `String`, then yes `TypeId` is what we want. 

We can do it in a separate PR as well, it's up to you 🙂

jansul: Definitely, this PR is big enough as it is 😂

src/ls/stack_reader.cr
```diff
@@ -0,0 +1,22 @@
+module Mint
+  module LS
+    class StackReader
+      def initialize(@stack : Array(Ast::Node))
+        @index = 0
+      end
+
+      def find_next(klass : N.class) : N | Nil forall N
```
Sija, line 8:
```suggestion
      def find_next(klass : N.class) : N? forall N
```

src/ls/stack_reader.cr
```diff
@@ -0,0 +1,22 @@
...
+        node = @stack[@index]?.try &.as?(N)
+        @index += 1
+        node
+      end
+
+      def find_anywhere(klass : N.class) : N | Nil forall N
```
Sija, line 14:
```suggestion
      def find_anywhere(klass : N.class) : N? forall N
```

Sija (approved):
I've left some minor comments but it looks awesome, great job! 🚀

PR #588, opened 2023-04-26
Refactor HTTP module a bit.

### LINKED ISSUE #494: Access to Http response headers
Hello, I'm rewriting a small application in Mint and have found that there's no way (as far as I can tell) to access the Http response headers when using the built-in Http module. For my use case, I need to know the content type of the response so I can decide how to parse it. 

I've run into this not just in my application but also when I'm using `mint start` during development; when I test my Http client requests, the Mint server will return 200 OK response to any request at any URL, and the response will contain the HTML from index.html. If the server always returns 200 OK and I have no way to access the content-type header, it seems I have no choice but to blindly attempt to decode it as several different formats until one succeeds. 

Thanks for your work on Mint! I'm really enjoying it as an alternative to React.

gdotdesign:
- Added response headers to `Http.Response` and `Http.ErrorResponse`
- Removed `Http.sendWithId`
- The second parameter of `Http.send` now has a default value
- The body of the response is now an enum and supports binary values

Should fix #494


core/source/Http.mint
```diff
@@ -297,25 +314,50 @@ module Http {
...
           status: `xhr.status`,
           url: request.url
         })})
       })
 
-      xhr.addEventListener('load', (event) => {
+      xhr.addEventListener('load', async (event) => {
         delete this._requests[#{uid}]
 
+        let header = xhr.getResponseHeader("content-type");
```
Sija, line 327:
```suggestion
        let contentType = xhr.getResponseHeader("Content-Type");
```

core/source/Http.mint
```diff
@@ -15,12 +15,23 @@ record Http.Request {
...
+  status : Number
+}
+
+/* Represents the body of an HTTP response. */
+enum Http.ResponseBody {
+  Json(Object)
+  Html(String)
+  Text(String)
+  File(File)
 }
```
Sija, line 30: We could also handle XML (via [`DOMParser`](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser))

```suggestion
enum Http.ResponseBody {
  Json(Object)
  XML(Object)
  Html(String)
  Text(String)
  File(File)
}
```

core/source/Http.mint
```diff
@@ -297,25 +315,72 @@ module Http {
...
           status: `xhr.status`,
           url: request.url
         })})
       })
 
-      xhr.addEventListener('load', (event) => {
+      xhr.addEventListener('load', async (event) => {
         delete this._requests[#{uid}]
 
+        let contentType = xhr.getResponseHeader("content-type");
```
Sija, line 328:
```suggestion
        let contentType = xhr.getResponseHeader("Content-Type");
```

gdotdesign: Nice catch :+1:

core/source/Http.mint
```diff
@@ -297,25 +315,72 @@ module Http {
...
+          try {
+            body = #{Http.ResponseBody::JSON(`JSON.parse(responseText)`)};
+          } catch (e) {
+            body = #{Http.ResponseBody::Text(`responseText`)};
+          }
+        } else if (contentType.startsWith("text/")) {
+          body = #{Http.ResponseBody::Text(`responseText`)};
+        }
+
+        if (!body) {
```
Sija, line 366: It seems to me this branch is impossible to fulfil, since `body` is always set beforehand?

core/source/Http.mint
Sija: It's purely stylistic, my ADHD-driven OCD couldn't not-notice it though :P

core/source/Http.mint
gdotdesign: The `body` is `null` to begin with https://github.com/mint-lang/mint/pull/588/files/9138b3d682ec1e763e9c3c22960d84dee6420cc9#diff-732cea90b2b18e39219ed170ca563d893caa3698fb7a515f8673748d4c473545R330 and there is no global `else` branch, so it can be `null` if we didn't match any types.

Sija: Also, shouldn't it turn the response body into a `File` only for `application/octet-stream` content-type responses?

core/source/Http.mint
gdotdesign: Oh, I was wrong, it's case-insensitive :joy: Well if it makes you happy :grin:

core/source/Http.mint
Sija: Oh, there's no `else` branch, I missed that 👍

gdotdesign: Well images, PDFs, document etc. should be files IMO, also the text version is always available in `bodyString`.

Sija (approved):

PR #589, opened 2023-04-27
Make parentheses optional for `case`, `for` and `if` expressions.

gdotdesign:
It can still parse parentheses but doesn't output them when formatting so basically deprecating it.

---

Background: I recently started playing with Rust, and they don't have parentheses for ifs and such, and I realized that they are not needed in Mint :slightly_smiling_face: 

- The changes to `.mint` files are from the formatter.


src/parsers/case.cr
```diff
@@ -13,7 +13,7 @@ module Mint
 
         whitespace
 
-        char '(', CaseExpectedOpeningParentheses
```
Sija, line 16: `CaseExpectedOpeningParentheses` can be removed, right?

src/parsers/if.cr
```diff
@@ -16,11 +16,11 @@ module Mint
         next unless keyword "if"
 
         whitespace
-        char '(', IfExpectedOpeningParentheses
```
Sija, line 19: ditto

src/parsers/case.cr
gdotdesign: Yep :+1:

Sija (approved):

PR #590, opened 2023-04-27
Allowing `let` statement in `if` expressions.

gdotdesign:


src/compilers/if.cr
```diff
@@ -25,7 +25,25 @@ module Mint
 
     def _compile(node : Ast::If, block : Proc(String, String)? = nil) : String
       condition =
-        compile node.condition
+        case statement = node.condition
+        when Ast::Statement
+          x =
```
Sija, line 30: Don't be lazy, please :P

```suggestion
          expression =
```

src/compilers/if.cr
```diff
@@ -50,7 +68,33 @@ module Mint
...
+          end
+
+        if condition[2]
+          js.asynciif do
+            js.statements([
+              condition[0],
+              js.return "(#{condition[1][0]} ? #{tru} : #{falsy})",
+            ])
+          end
+        else
```
Sija, line 87:
```suggestion
        truthy =
          js.iif do
            js.statements(condition[1][1] + [
              js.return truthy,
            ])
          end

        if condition[2]
          js.asynciif do
            js.statements([
              condition[0],
              js.return "(#{condition[1][0]} ? #{truthy} : #{falsy})",
            ])
          end
        else
```

src/formatters/if.cr
```diff
@@ -2,7 +2,12 @@ module Mint
   class Formatter
     def format(node : Ast::If) : String
       condition =
-        format node.condition
+        case x = node.condition
+        when Ast::Statement
+          format x, false
```
Sija, line 7:
```suggestion
        case condition = node.condition
        when Ast::Statement
          format condition, false
```

Sija (changes requested):

src/compilers/if.cr
gdotdesign: Sorry, I created the PR and then converted to draft :joy:

src/compilers/if.cr
gdotdesign: There is already a `truthy` variable earlier in the file, don't know yet what to call this :shrug:

src/formatters/if.cr
```diff
@@ -2,7 +2,12 @@ module Mint
   class Formatter
     def format(node : Ast::If) : String
       condition =
-        format node.condition
+        case item = node.condition
+        when Ast::Statement
+          format item, false
+        else
+          format node.condition
```
Sija, line 9:
```suggestion
          format item
```

Sija (approved):
🚀

spec/compilers/if_let
```diff
@@ -0,0 +1,44 @@
+enum T {
+  A(String)
+  B
+}
+
+component Main {
+  fun render : String {
+    if (let T::A(a) = T::A("")) {
```
Sija, line 8: Spec for assignment without parens would be a nice thing to have.

PR #592, opened 2023-05-01
Convert type_id from String to a node

jansul:
This PR modifies the `type_id`/`type_id!` methods of the parser to return a new node (`Ast::TypeId`) rather than return a `String`. Any other node affected by this has also been updated.

This should make it easier to implement language server features where it can be useful to exact positions of some nodes. e.g. the `name` of an `Ast::Component`

I've tested this change against the `mint-realworld` repository, running `mint docs` etc and everything seemed OK, but could possible use more testing given how much is changed.

### DIFF
--- a/src/ls/hover.cr
+++ b/src/ls/hover.cr
@@ -60,8 +60,8 @@ module Mint
             case node
             when Ast::Variable, Ast::TypeId
               # If the first node under the cursor is a `Ast::Variable`
-              # or `LSP::TypeId`, then get the associated nodes 
-              # information and hover that otherwise get the hover 
+              # or `Ast::TypeId`, then get the associated nodes
+              # information and hover that otherwise get the hover
               # information of the parent.
               lookup =
                 workspace.type_checker.variables[node]?
--- a/src/ls/hover.cr
+++ b/src/ls/hover.cr
@@ -58,10 +58,11 @@ module Mint
             parent = stack[1]?
 
             case node
-            when Ast::Variable
-              # If the first node under the cursor is a variable then
-              # get the associated nodes information and hover that
-              # otherwise get the hover information of the parent.
+            when Ast::Variable, Ast::TypeId
+              # If the first node under the cursor is a `Ast::Variable`
+              # or `LSP::TypeId`, then get the associated nodes 
+              # information and hover that otherwise get the hover 
+              # information of the parent.
               lookup =
                 workspace.type_checker.variables[node]?
 
--- a/src/parsers/html_body.cr
+++ b/src/parsers/html_body.cr
@@ -38,9 +38,7 @@ module Mint
 
         closing_tag =
           case tag
-          when Ast::Variable
-            tag.value
-          when Ast::TypeId
+          when Ast::Variable, Ast::TypeId
             tag.value
           else
             tag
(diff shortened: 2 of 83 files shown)


src/ls/hover.cr
```diff
@@ -58,10 +58,11 @@ module Mint
             parent = stack[1]?
 
             case node
-            when Ast::Variable
-              # If the first node under the cursor is a variable then
-              # get the associated nodes information and hover that
-              # otherwise get the hover information of the parent.
+            when Ast::Variable, Ast::TypeId
+              # If the first node under the cursor is a `Ast::Variable`
+              # or `LSP::TypeId`, then get the associated nodes
```
gdotdesign, line 63: It's not `Ast::TypeId`?

gdotdesign (approved):
Looks good :+1: I guess if it compiles and works with the real word repo it should be good! Once merged I'll test `mint-ui` with it.

src/parsers/html_body.cr
```diff
@@ -40,6 +40,8 @@ module Mint
           case tag
           when Ast::Variable
             tag.value
+          when Ast::TypeId
+            tag.value
```
Sija, line 44:
```suggestion
          when Ast::Variable, Ast::TypeId
            tag.value
```

src/type_checkers/use.cr
```diff
@@ -22,7 +22,7 @@ module Mint
 
       # This is checked by the provider so we assume it's there
       subscription =
-        records.find!(&.name.==(provider.subscription))
+        records.find!(&.name.==(provider.subscription.value))
```
Sija, line 25: `#name` is not a `TypeId`?

```suggestion
        records.find!(&.name.value.==(provider.subscription.value))
```

Sija (commented):
I'm wondering, wouldn't be cleaner to implement `TypeId#to_s` and `Comparable(String)` to avoid explicit usage of `#value` property?

gdotdesign:
> I'm wondering, wouldn't be cleaner to implement `TypeId#to_s` and `Comparable(String)` to avoid explicit usage of `#value` property?

I guess we could do that, however we currently don't, so I don't know if it's in the scope of this PR or not :thinking:

Sija:
@gdotdesign I'm pretty sure we could at least implement `TypeId#==(other: TypeId)` to avoid doin' in manually everywhere, like:

```cr
ast.stores.find(&.name.value.==(connect.store.value))
```

src/type_checkers/use.cr
jansul: Actually not in this case :smile:!  `records` here is a list of `TypeChecker::Record`'s, where `name` is a [`String`](https://github.com/mint-lang/mint/blob/master/src/type_checker/record.cr#L8)

jansul:
I was very tempted to implement those methods, however as it's not something we are doing on any other node (especially with `Ast::Variable` where we also use `.value`) I thought best to do it like this for consistency.

Happy to make any changes :smile:  `TypeId#==(other: TypeId)` would probably be the safest and least confusing. 

Just noticed a bug where I'm not using `.value` when passing to errors so I will fix that too ...!
image

jansul:
Fixed all of the above :smile:  

As an aside it may be worth hardening the [`bold` method of `BlockBuilder`](https://github.com/mint-lang/mint/blob/master/src/message.cr#L19) to only accept strings the same as `text` as that would have caught that issue!

Sija (approved):

PR #593, opened 2023-05-05
Nested Destructuring

gdotdesign:
This implements a recursive algorithm for destructuring and unifies statement and case branch destructuring checks. This allows:

- Destructuring with values in them
- Mixing and nesting destucturings
- Using complex destructurings in `if` expressions

Some examples:

```mint
case Test::Record(name: Maybe::Just({"John", "Doe"}), age: 32) {
  Test::Record(Maybe::Just({"John", a})) => a
  => "WTF"
}
    
case Result::Ok(Maybe::Just({"Hello", "Bello"})) {
  Result::Ok(Maybe::Just({"Hello", b})) => "Matches"
  Result::Ok(Maybe::Just({"JULIE"})) => "Not matches"
  Result::Err => "Error"
  => "Else branch"
}

case ["a", "b", "d"] {
  [a, ...b, d] => b
  ["a"] => ["one"]
  [a] => ["something else"]
  [] => ["empty"]
}

if Result::Ok(Maybe::Just({"Hello", b})) = result {
  ...
} else {
  ...
}
```

Top level exhaustiveness checking is still there but if any branch has nested destructuring or value (`"string"` for example) in it then it's excluded from the check.

Follow-up work (in separate PRs):
- Move `_match` and related functions into the runtime
- Allow named destructuring of records (including in enums) 
- At some point implement exhaustiveness checking in sublevels


core/source/Http.mint
```diff
@@ -334,7 +334,7 @@ module Http {
             (new DOMParser()).parseFromString(responseText, "text/html");
 
           const errorNode =
-            doc.querySelector("parsererror");
+            object.querySelector("parsererror");
```
gdotdesign, line 337: This is an unrelated fix which will probably go into a different commit.

spec/compilers/case
```diff
@@ -1,36 +1,38 @@
 component Main {
-  fun test : Bool {
+  fun render : String {
```
Sija, line 2:
```suggestion
  fun render : Bool {
```

src/parsers/variable.cr
```diff
@@ -45,7 +45,7 @@ module Mint
       variable(track) || raise error
     end
 
-    def variable(track = true, invalid = [] of String) : Ast::Variable?
+    def variable(track = true, invalid = ["true", "false"]) : Ast::Variable?
```
Sija, line 48: I'd keep the constant for performance reasons.

```suggestion
    def variable(track = true, invalid = INVALID_VARIABLE_NAMES) : Ast::Variable?
```

gdotdesign:
> @gdotdesign Does it resolves #283?

Yes it will :)

spec/compilers/case
gdotdesign: String is the return type of this function from the value below the `case`.

src/parsers/variable.cr
gdotdesign: Moved it from the arguments into the method body.

src/compilers/top_level.cr
```diff
@@ -266,6 +266,113 @@ module Mint
         const _S = mint.Store
         const _E = mint.Enum
 
+        const _PR = (patterns) => new RecordPattern(patterns)
+        const _PE = (x, pattern) => new Pattern(x, pattern)
+        const _PV = Symbol("Variable");
+        const _PS = Symbol("Sperad");
```
Sija, line 272:
```suggestion
        const _PV = Symbol("Variable")
        const _PS = Symbol("Spread")
```

src/parsers/variable.cr
```diff
@@ -45,15 +45,15 @@ module Mint
...
+    def variable(track = true) : Ast::Variable?
       start do |start_position|
         value = gather do
           next unless char.ascii_lowercase?
           letters_or_numbers
         end
 
         next unless value
-        next if value.in?(invalid)
+        next if value.in?(%w[true false])
```
Sija, line 56: Please, move it into a constant to avoid redundant allocations.

src/type_checkers/case.cr
```diff
@@ -24,7 +24,7 @@ module Mint
 
       unified =
         node
-          .branches[1..node.branches.size]
+          .branches[(1..)]
```
Sija, line 27:
```suggestion
          .branches[1..]
```

src/type_checkers/case.cr
```diff
@@ -84,7 +89,15 @@ module Mint
           } if !not_matched.empty? && !catch_all
         elsif condition.name == "Array"
           destructurings =
-            node.branches.map(&.match).select(Ast::ArrayDestructuring)
+            node.branches
+              .map(&.match)
+              .select(Ast::ArrayDestructuring)
+              .select do |branch|
```
Sija, line 95:
```suggestion
              .select! do |branch|
```

src/type_checkers/case.cr
```diff
@@ -50,19 +50,24 @@ module Mint
...
-        if item
+        if parent
           not_matched =
-            item.options.reject do |option|
+            parent.options.reject do |option|
               node
                 .branches
                 .compact_map(&.match)
                 .any? do |match|
                   case match
```
Sija, line 63:
```suggestion
              node
                .branches
                .any? do |branch|
                  case match = branch.match
```

PR #594, opened 2023-05-05
Update `Ast::Constant` to use an `Ast::Variable` for its name

jansul:
This PR updates `Ast::Constant` to use an `Ast::Variable` for its name, rather than a string. This will be useful for future changes to the language server.

Unfortunately, similar to the recent `Ast::TypeId` change, it includes a lot of `.value`'s, although there were a couple of places where I didn't seem to need to add it such as in the formatter.

This PR also fixes a bug where some invalid constants were allowed, notably ones that start with a number or underscore. (e.g `_CONSTANT`, `1CONSTANT`)

### DIFF
--- a/src/compilers/variable.cr
+++ b/src/compilers/variable.cr
@@ -20,7 +20,7 @@ module Mint
             when Ast::Function then entity.name.value
             when Ast::State    then entity.name.value
             when Ast::Get      then entity.name.value
-            when Ast::Constant then entity.name
+            when Ast::Constant then entity.name.value
             end
 
           if store
--- a/src/parsers/variable.cr
+++ b/src/parsers/variable.cr
@@ -41,6 +41,26 @@ module Mint
       end
     end
 
+    def variable_constant! : Ast::Variable
+      start do |start_position|
+        head =
+          gather { chars &.ascii_uppercase? }
+
+        tail =
+          gather { chars { |char| char.ascii_uppercase? || char.ascii_number? || char == '_' } }
+
+        raise ConstantExpectedName unless head
+
+        value = "#{head}#{tail}"
+
+        Ast::Variable.new(
+          from: start_position,
+          value: value,
+          to: position,
+          input: data)
+      end
+    end
+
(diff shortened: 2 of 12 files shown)


src/compilers/variable.cr
```diff
@@ -20,7 +20,7 @@ module Mint
             when Ast::Function then entity.name.value
             when Ast::State    then entity.name.value
             when Ast::Get      then entity.name.value
-            when Ast::Constant then entity.name
+            when Ast::Constant then entity.name.value
```
Sija, line 23: These branches could be unified via `..., ...` notation.

Sija (approved):

src/parsers/variable.cr
```diff
@@ -41,6 +41,26 @@ module Mint
       end
     end
 
+    def variable_constant! : Ast::Variable
```
Sija, line 44: Shouldn't this be called?

```suggestion
    def constant! : Ast::Variable
```

jansul: I prefixed it with `variable_` to be consistent with some of the other methods in this file (i.e. `variable_attribute_name`, `variable_with_dashes`) but I do prefer your suggestion.

Happy to change it if you like?

Sija: Some cases are legitimate like `variable_with_dashes`, yet in others (like this one, or `variable_attribute_name`) the `variable_` prefix seems like an implementation detail, and thus redundant, wdyt @gdotdesign ?

gdotdesign: We have `constant` which returns `Ast::Constant?` so by convention if it would be called `constant!` it should return `Ast::Constant` and raise when it could not parse, so I guess the prefix is fine.

gdotdesign (approved):

src/parsers/variable.cr
Sija: Oh, I see... In fact it's a constant _name_ in the end?

gdotdesign: Yes.

PR #595, opened 2023-05-05
LSP: Refactor definition to be similar to hover feature

jansul:
Unfortunately it's a little hard to see the actual change in this PR between all the red and green, so hopefully some description helps.

After understanding how the hover implementation works a little better, I've refactored the definition code to be more in line with it. 

---

The biggest change is that I have reverted which nodes we track (see `track: false` everywhere), so we are no longer tracking more specific nodes such as component names etc

Instead,  there is a new method `cursor_intersects?` which can be used to check if a node is currently being hovered over. We can use this to exit early when trying to find the definition for a "higher level" node.

e.g.
```crystal
def definition(node : Ast::HtmlComponent, server : Server, workspace : Workspace, stack : Array(Ast::Node))
  
  # We are not hovering over the component name, so don't bother looking for the definition
  return unless cursor_intersects?(node.component)

  # ...
end
```

This removes the need for `StackReader`, as that was mostly used to find the reverse ("higher level" nodes from more specific ones)

---

I've implemented some more of the easier "Go to Definitions" (for a separate PR for ease of review!) and it doesn't seem any more difficult to implement using this method.


Sija (approved):
Looks awesome 🚀

src/ls/definition.cr
```diff
@@ -15,84 +15,30 @@ module Mint
...
-        selection(node.name)
-      end
-
-      def selection(node : Ast::HtmlComponent) : LSP::Range
-        # Select only the name part of the component
-        # <Component attribute={value}>
-        #  ^^^^^^^^^
-        selection(node.component)
+      def cursor_intersects?(node : Ast::Node, position : LSP::Position) : Bool
+        node.location.contains?(position.line + 1, position.character)
```
gdotdesign, line 29: The tests are passing, so it must be right just curious why `+ 1`?

src/ls/definition/html_element.cr
```diff
@@ -0,0 +1,19 @@
...
+  module LS
+    class Definition < LSP::RequestMessage
+      def definition(node : Ast::HtmlElement, server : Server, workspace : Workspace, stack : Array(Ast::Node))
+        node.styles.each do |style|
+          next unless cursor_intersects?(style)
+
+          return definition(style, server, workspace, stack)
+        end
+
+        node.attributes.each do |attribute|
```
gdotdesign, line 11: I think this is unnecessary since attributes for HTML elements are not defined anywhere for example `onClick` or `method` for `form`, but maybe I'm missing something :slightly_smiling_face:

src/ls/definition.cr
jansul: The real answer is that I just copied it from [`nodes_at_cursor`](https://github.com/mint-lang/mint/blob/master/src/ls/server.cr#L37) ... :sweat_smile:  

But also, it's because for `node.location`'s the line number [is 1-based ](https://github.com/mint-lang/mint/blob/master/src/ast/node.cr#L103), but in the language server specification both line and character are [zero based](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position)

src/ls/definition/html_element.cr
jansul: Yes you are completely correct, only `styles` needs to be checked here.

It's actually doubly redundant as `Ast::HtmlAttribute` is a [tracked node](https://github.com/mint-lang/mint/blob/master/src/parsers/html_attribute.cr#L31), which is why it still works for `Ast::HtmlComponent` without any similar logic. :sweat_smile:

src/ls/definition.cr
gdotdesign: Yeah, I had a feeling 😅 thanks for confirming 🙏

gdotdesign (approved):

PR #597, opened 2023-05-09
LSP: Add Go to Definition support for Enum/Type/Connect

jansul:
This PR adds "Go to Definition" support to some more of the easier nodes as shown in the videos below.

For `Ast::ConnectVariable`:

https://github.com/mint-lang/mint/assets/1927518/64ab6659-c07a-442b-b70b-adec9698ada6

For `Ast::EnumDestructuring`:

https://github.com/mint-lang/mint/assets/1927518/2b372165-a1be-446a-ae77-44809c5fd6a8

For `Ast::EnumId`:

https://github.com/mint-lang/mint/assets/1927518/2278adaf-8013-4cbd-9f0f-92746926a727

For `Ast::Type`:

https://github.com/mint-lang/mint/assets/1927518/4b181727-c4e0-4165-9fa5-d2576fa17066

There's still some testing I need to do, especially when referencing any of the "core" files (they are not on the filesystem so you cannot link to them :cry: )  but otherwise should be good for review.

### DIFF
--- a/spec/language_server/definition/core_component
+++ b/spec/language_server/definition/core_component
@@ -4,6 +4,21 @@ component Test {
       <div/>
     </Unless>
   }
+
+  fun aborted : Http.Error {
+    Http.Error::Aborted
+  }
+
+  state error : Http.Error = Http.Error::Aborted
+
+  fun toString (error : Http.Error) : String {
+    case error {
+      Http.Error::Aborted => "Aborted"
+      Http.Error::NetworkError => "NetworkError"
+      Http.Error::Timeout => "Timeout"
+      Http.Error::BadUrl => "BadUrl"
+    }
+  }
 }
 ------------------------------------------------------------------file test.mint
 {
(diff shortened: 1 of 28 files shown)


jansul:
Somewhat related, I think I might be able to solve linking to core files (components, modules etc) without them having to be on  disk using this:
https://code.visualstudio.com/api/extension-guides/virtual-documents

Unfortunately would be VSCode specific until the LSP itself supports similar functionality

gdotdesign:
Awesome progress :tada:  These are very useful to have :pray: 

> There's still some testing I need to do, especially when referencing any of the "core" files (they are not on the filesystem so you cannot link to them cry ) but otherwise should be good for review.

We could show the source of the definition in a message https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessageRequest until something like virtual documents are implemented, or we can dump them somewhere (`.mint` directory) and link to them maybe?

gdotdesign (approved):

jansul:
> or we can dump them somewhere (`.mint` directory)

I had thought of this, but it seemed wrong spamming the whole standard library to disk just for this feature - as useful as it would be. (getting flashbacks of `node_modules` 😱 )

Looks like there are discussions [about adding something similar to virtual documents to the LS specification](https://github.com/microsoft/language-server-protocol/issues/1264), so hopefully something that isn't VSCode specific wouldn't be far off.

In the meantime I will probably just make it so you can't link to core files, or show a message as mentioned.

Sija (approved):

jansul:
Fixed issues when linking to core files. At the moment it just returns an empty definition response until we can implement something like the above :smile:  

Noticed a small bug when testing it:  

Screenshot 2023-05-11 at 20-04-20 Screenshot

Will fix in another PR. Might try and add a few tests for `messages/` in general as I'm sure my `TypeId` etc changes has probably broken a few others :sweat_smile:

PR #598, opened 2023-05-11
Allow `else` branch to be omitted in certain cases.

gdotdesign:
This PR contains three separate changes (sorry about that):

1. Allow omitting the `else` branch when the return type and value can be inferred. The reason for this is there are cases when the else branch is unnecessary like when working with `Html` (which could be omitted before) or `next { }` branches. The inferred return values for omitted else branches are:
	- `String` returns empty string `""`
	- `Array(a)` returns empty array
	- `Promise(Void)` returns `null` (maybe it will need to return an actual promise we will see)
	- `Maybe(a)` returns `Maybe::Nothing`
	- `Void` returns `null`
	- `Html` returns `null`
2. `next { }` now compiles to `null`. Previously it would compile to something like this ( which essentially does nothing):
   ```
   new Promise((resolve) => { this.setState(new Record({})) })
   ```
3. Remove `NEVER` from the type checker and replace it with `VOID`

### DIFF
--- a/spec/compilers/if_without_else_array
+++ b/spec/compilers/if_without_else_array
@@ -1,0 +1,15 @@
+component Main {
+  fun render : Array(String) {
+    if ("asd" == "asd2") {
+      ["ARRAY"]
+    }
+  }
+}
+--------------------------------------------------------------------------------
+class A extends _C {
+  render() {
+    return (_compare(`asd`, `asd2`) ? [`ARRAY`] : []);
+  }
+};
+
+A.displayName = "Main";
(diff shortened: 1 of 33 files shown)


jansul:
`String` might be the only one that's a little too magic for my taste 😅  (Can always use `Maybe(String)` for an optional string?)

Do you have any use cases for that one in mind?

gdotdesign:
> `String` might be the only one that's a little too magic for my taste sweat_smile (Can always use `Maybe(String)` for an optional string?)
> 
> Do you have any use cases for that one in mind?

It's mainly for `Html` (as it currently works):

```
<div>
  if condition {
    "value"
  }
</div>
```

Sija (approved):

PR #599, opened 2023-05-11
Fix comment being included in `case_enum_not_covered` message

jansul:
This fixes the issue mentioned [here](https://github.com/mint-lang/mint/pull/597#issuecomment-1544562561). Although opening as a draft to see if everyone is happy with this method of testing messages.

As most of the exceptions covered by the messages already have test cases (see `spec/parsers` and `spec/type_checking`), it should be possible to generate these tests with not too much effort - although not this one specifically, as the original test didn't have a comment!

Only thing is that it would be a lot of files :sweat_smile:

### DIFF
--- a/src/type_checkers/case.cr
+++ b/src/type_checkers/case.cr
@@ -79,7 +79,7 @@ module Mint
 
           options =
             not_matched.map do |option|
-              "#{format parent.name}::#{formatter.replace_skipped(format(option))}"
+              "#{format parent.name}::#{formatter.replace_skipped(format(option.value))}"
             end
 
           raise CaseEnumNotCovered, {


Sija:
I'd try to avoid goin' into style-specifics (i.e. the literal output), since once these change (and they will at some point), all of these tests would need to be regenerated.

Sija:
Perhaps, let's merge this fix and leave tests alone in this PR and create a separate one just for this.

Sija (approved):

jansul:
I'll create an issue for discussion instead maybe? 

Not the highest priority tbh, just might be nice to have when doing larger refactors (saw a few error messages containing `#<Mint::Ast::TypeId...` while making that change, worried I broke more :sweat_smile: )

Sija:
@jansul Great idea! Yeah, tests would be much welcomed :)

Sija:
Is this GTG?

jansul:
> Is this GTG?

Yep! Here's how the error is display after this change

image

PR #601, opened 2023-05-18
Early return.

gdotdesign:
This PR adds the ability to return early from a block using the `return` statement.

----

Since the removing the `sequence` and `try` keywords the only way to handle errors was with a case statement, which leads to nested `case` expressions like so:

```
fun load (localePath : String, cheatsPath : String) {
    if (cacheKeys != {localePath, cheatsPath}) {
      case (await Http.send(Http.get(localePath))) {
        Result::Err => next { }

        Result::Ok(response) =>
          case (Json.parse(response.bodyString)) {
            Result::Err => next { }

            Result::Ok(object) =>
              case (decode object as CheatSheet) {
                Result::Err => next { }

                Result::Ok(sheet) =>
                  case (await Http.send(Http.get(cheatsPath))) {
                    Result::Err => next { }

                    Result::Ok(response1) =>
                      case (Json.parse(response1.bodyString)) {
                        Result::Err => next { }

                        Result::Ok(object1) =>
                          case (decode object1 as Array(Array(String))) {
                            Result::Err => next { }

                            Result::Ok(values) =>
                              next
                                {
                                  cacheKeys: {localePath, cheatsPath},
                                  values: values,
                                  sheet: sheet
                                }
                          }
                      }
                  }
              }
          }
      }
    } else {
      Promise.never()
    }
  }
```

Using early return results in this:

```
fun load (localePath : String, cheatsPath : String) {
  if (cacheKeys != {localePath, cheatsPath}) {
    let Result::Ok(response) =
      await Http.send(Http.get(localePath)) or return next { }

    let Result::Ok(object) =
      Json.parse(response.bodyString) or return next { }

    let Result::Ok(sheet) =
      decode object as CheatSheet or return next { }

    let Result::Ok(response1) =
      await Http.send(Http.get(cheatsPath)) or return next { }

    let Result::Ok(object1) =
      Json.parse(response1.bodyString) or return void

    let Result::Ok(values) =
      decode object1 as Array(Array(String)) or return next { }

    next
      {
        cacheKeys: {localePath, cheatsPath},
        values: values,
        sheet: sheet
      }
  } else {
    next { }
  }
}
```

There are a few rules for early return so far:

- It's scoped to the block not the function
- It can only appear as a standalone statement or as the right operand of an `or` operation with the left operand being a destructuring which is not exhaustive

----

Changes in this PR:

- Added `return` statement.
- Refactored how statements are resolved during type checking. They could reference each other no matter in which the order they were. Now they are resolved sequentially.
- Using a `let` as the last statement of a block now results in a type error.
- It is now possible to destructure any type not just tuples in statemtents. If the destructuring is not exhaustive it requires a return call.


jansul:
This is cool !

Tried to think of some other syntax like `?` or `??` operators but think as it is might be the cleanest. 😄 

What happens when you omit the `or return void`? Or is it always required?

e.g.
```mint
{
   let Result::Ok(response) =
      await Http.send(Http.get(localePath)))
}

```

gdotdesign:
> Tried to think of some other syntax like `?` or `??` operators but think as it is might be the cleanest. smile

I thought about this a lot :thinking: and researched how other languages do it:

- Ruby, JavaScript, Crystal, etc... have exceptions which is not ideal because it can bubble up, and they can be forgotten resulting in runtime errors
- Haskell has `do` notation with defaults for types which are implemented in a different location, so you might not know what the default is
- Rust has `?` which is error propagation, but it only works with results (and the error type needs to be the same) and `unwrap` which panics

This is the cleanest and easier to reason about way that I could come up with :slightly_smiling_face: I might write a blog post about this sometime :thinking: 

> What happens when you omit the or return void? Or is it always required?

It's required when the destructuring is not exhaustive because the other possibilities need to be handled (e.g. what happens when the request fails?)

`or return x` is basically a shorthand for `if` but with the difference of ending the execution of the block early:
```
let Result::Ok(response) =
  await Http.send(Http.get(localePath)) or return void

if let Result::Ok(response) = await Http.send(Http.get(localePath))) {
  stuff
} else {
  void
}
```

Sija (approved):

jansul:
Makes sense, thanks!

> This is the cleanest and easier to reason about way that I could come up with 🙂 I might write a blog post about this sometime 🤔

I've also been thinking about it and can't think of anything better 😅 

I do have one question, is `return` outside of using it with `or return` also scoped to the block, or is that scoped to function? I could see some confusion if that was different

PR #603, opened 2023-05-24
Add `Url.toString()` API function

nilslindemann:
[Number](https://github.com/mint-lang/mint/blob/26c370f786ea56e66d12d23df804304e506efa01/core/source/Number.mint#L103-L105), [Bool](https://github.com/mint-lang/mint/blob/26c370f786ea56e66d12d23df804304e506efa01/core/source/Bool.mint#L9-L11), [Regexp](https://github.com/mint-lang/mint/blob/26c370f786ea56e66d12d23df804304e506efa01/core/source/Regexp.mint#L194-L196) have it too.

### DIFF
--- a/core/source/Url.mint
+++ b/core/source/Url.mint
@@ -67,6 +67,16 @@ module Url {
   }
 
   /*
+  Converts the given `Url` to a `String`.
+
+    (Url.parse("https://www.example.com/path/?search=foo#hash")
+    |> Url.toString()) == "https://www.example.com/path/?search=foo#hash"
+  */
+  fun toString (url : Url) : String {
+    `#{url}.origin + #{url}.path + #{url}.search + #{url}.hash`
+  }
+
+  /*
   Releases an existing object URL which was previously created.
 
     Url.createObjectUrlFromString("Content", "text/html")
--- a/core/tests/tests/Url.mint
+++ b/core/tests/tests/Url.mint
@@ -8,6 +8,13 @@ suite "Url.parse" {
   }
 }
 
+suite "Url.toString" {
+  test "converts an url to a string" {
+    (Url.parse("https://www.example.com/path/?search=foo#hash")
+    |> Url.toString()) == "https://www.example.com/path/?search=foo#hash"
+  }
+}
+
 suite "Url.createObjectUrlFromFile" {
   test "it creates an url from a file" {
     (File.fromString("Content", "test.html", "text/html")


core/source/Url.mint
```diff
@@ -66,6 +66,16 @@ module Url {
...
   }
 
+  /*
+  Converts the given `Url` to a `String`.
+
+    (Url.parse("https://www.example.com/path/?search=foo#hash")
+    |> Url.toString()) == "https://www.example.com/path/?search=foo#hash"
+  */
+  fun toString (url : Url) : String {
+    `#{url}.origin + #{url}.path + #{url}.search + #{url}.hash`
```
Sija, line 76: Why not use built in `toString()`?

```suggestion
    `#{url}.toString()`
```

nilslindemann: `Url` [is an object internally](https://github.com/mint-lang/mint/blob/26c370f786ea56e66d12d23df804304e506efa01/core/source/Url.mint#L55-L64), so calling `.toString()` on it would return `"[object Object]"`.

nilslindemann:
Sorry, I do not know yet how to fix these failed checks.

core/source/Url.mint
nilslindemann: I would guess, to return `"[object Object]"` would be the task of an `.repr` function, comparable to `repr` in Python. But even then, I wonder if `"[object Object]"` is a useful representation. Javascript has done this not so userfriendly in my opinion. Something like `JSON.stringify(url)` would have fitted more.

Sija: Following code prints the url back 🤷‍♂️ 

```js
const url = new URL("https://github.com/mint-lang/mint/pull/603#discussion_r1204579853")
console.log(url.toString()) // => https://github.com/mint-lang/mint/pull/603#discussion_r1204579853
```

Sija: Oh, right. The mint's `Url` is not backed by the JS' `URL`, my bad.

gdotdesign: > @gdotdesign Was the decision to decouple the `Url` from the JS counterpart was done on purpose?

Yes, `Url` in Mint is a record. Using the JS `URL` would mean it would be an opaque type and helper methods would have been needed for getting it's parts.

gdotdesign:
> Sorry, I do not know yet how to fix these failed checks.

Those are unrelated to the PR itself so no worries :)

gdotdesign (approved):

PR #611, opened 2023-05-28
When no hash, jump to start of document

nilslindemann:
When the user calls `Window.triggerHashJump()`, and there is no hash part in the URL, jump to the start of the document.


gdotdesign (approved):

nilslindemann:
thx

PR #612, opened 2023-05-30
Improve number handling

nilslindemann:


nilslindemann:
Oops, `Number` test fails, I am on it.

core/source/Number.mint
```diff
@@ -42,9 +42,13 @@ module Number {
...
     `
     (() => {
-      let value = parseFloat(#{input})
+      let value = Number(#{input})
 
-      if (isNaN(value)) {
+      if (
+        Number.isNaN(value)
+        ||
+        Number.isNaN(value - parseFloat(#{input}))  // excludes "Infinity", "-Infinity", ""
```
gdotdesign, line 50: Why not compare with `Infinity` and `-Infinity`?

core/source/Number.mint
```diff
@@ -42,9 +42,13 @@ module Number {
   fun fromString (input : String) : Maybe(Number) {
     `
     (() => {
-      let value = parseFloat(#{input})
+      let value = Number(#{input})
```
gdotdesign, line 45: This seems like a breaking change (from this https://stackoverflow.com/questions/11988547/what-is-the-difference-between-number-and-parsefloat):
```
parseFloat('3.23abc'); //=> 3.23
Number('3.23abc'); //=> NaN
```

Not sure which one is better though :man_shrugging: maybe try parsing `Number` first and if it fails fallback to `parseFloat`? @Sija any opinions?

core/tests/tests/Validation.mint
```diff
@@ -11,11 +11,33 @@ suite "Validation.isNotBlank" {
 suite "Validation.isNumber" {
   test "it returns nothing if the string can be converted to number" {
     Validation.isNumber("0", {"key", "ERROR"}) == Maybe::Nothing
+    Validation.isNumber("0.1", {"key", "ERROR"}) == Maybe::Nothing
```
gdotdesign, line 14: Line 13 and 14 will not count in the assertion. They need to be chained with `&&` or put into separate test cases.

core/tests/tests/Validation.mint
```diff
@@ -11,11 +11,33 @@ suite "Validation.isNotBlank" {
...
+  }
+
+  test "it returns the error if the string is empty" {
+    Validation.isDigits("", {"key", "ERROR"}) == Maybe::Just({"key", "ERROR"})
+  }
+
+  test "it returns the error if the string contains non-digits" {
+    Validation.isDigits("a", {"key", "ERROR"}) == Maybe::Just({"key", "ERROR"})
+    Validation.isDigits(" 1 ", {"key", "ERROR"}) == Maybe::Just({"key", "ERROR"})
+    Validation.isDigits("1.2", {"key", "ERROR"}) == Maybe::Just({"key", "ERROR"})
```
gdotdesign, line 39: Same as above with lines 37 and 38.

core/source/Number.mint
nilslindemann: I was actually asking this myself too. I am open to including it, but I am unsure about parsing `""` as `0`. It may result in problems when doing `Validate.isNumber()`.

core/source/Number.mint
nilslindemann: Changing this was actually the main motivation behind this PR. Without this reduction, `Validate.isNumber()` will validate when a zip input has the value `1234x`. Someone may accidentally mistype his age as `8p`. It gets accepted and suddenly he is 8 instead of 80.

core/tests/tests/Validation.mint
nilslindemann: Thank you, I was not aware of that.

core/source/Number.mint
Sija: Why not leave `parseFloat`?

core/source/Number.mint
gdotdesign: I think we need to split this function into two with appropriate documentation:

1. `Number.fromString` should be strict, using `Number` and return `Maybe::Nothing` instead of `0` for empty string
	
2. `Number.fromStringRelaxed` should be relaxed, using `parseFloat`

Also, possibly return `Maybe::Nothing` if someone passes a non `String` value (in case of inline JS) for both.

Another option would be to have a second argument with a default value but I don't like that much.

core/source/Number.mint
gdotdesign: @nilslindemann provided a reason in another comment https://github.com/mint-lang/mint/pull/612#discussion_r1210172933, but basically `8O` would be `8` instead of `80` thus can cause confusion.

core/source/Number.mint
Sija: All good points IMO, I'd go for `Number.fromString(string, strict: true)` signature though, since it's essentially the same behaviour, just varied by the amount of validation.

nilslindemann: > All good points IMO, I'd go for Number.fromString(string, strict: true) signature though, since it's essentially the same behaviour, just varied by the amount of validation.

It is a pity, though, that we can not could call it like this: `Number.fromString("1", strict=true)`. 

> TIL: nifty trick to do a strict number (float) conversion

It can be done [even shorter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unary_plus): `+"x-1.11"` ;-)

gdotdesign: > It is a pity, though, that we can not could call it like this: `Number.fromString("1", strict=true)`.

I'm planning to implement something like that 😉 (would be `Number.fromString("1", strict: true)` to match other things).

We can go with the second argument if you both think it's better :+1:

nilslindemann: I will do.

Sija: - I'd say it should be added to the language, perhaps as a const (`Number.Infinity`)?
- Good point
- I'd say it's either this, or error

gdotdesign: * Yes, a constant for it makes sense.
* That will need to change to return `Result(Number.Error, String)`
* It's consistent with JavaScript so I guess it can stay as is.
* `Number.toFixed` needs to return a result as well since it can throw a `RangeError` `Infinity.toFixed(200)`

nilslindemann: Ok, but it is a lot of work, basically, go through the whole library and check, and the point of this PR was not to add Infinity to Mint (It did not exist before), but to fix the cases when `Validate.isNumber()` shows a weird behavior.

Further, If the addition of Infinity leads to code being more bloated, like, `Number.format` returns `Result(Number.Error, String)` instead of just `String`, then this is a big problem in my opinion. Give me some practical use cases for Infinity, that justify this overhead for library users.

nilslindemann: Ok, I just see, parseFloat(Infinity) actually returns Infinity. So this was an unhandled edge case before.

nilslindemann: I will handle it in this function, but keep the other functions as they were for now.

nilslindemann:
That could have been "Include Infinity, allow only strings _as input_".

nilslindemann:
As a side effect, it now runs approximately five times faster (`Number()` itself is around 2000 times faster than `parseFloat()`)

core/source/Number.mint
```diff
@@ -42,17 +42,21 @@ module Number {
   fun fromString (input : String) : Maybe(Number) {
     `
     (() => {
+      if (typeof #{input} !== 'string') {
+        return #{Maybe::Nothing}
+      }
+
```
Sija: `input` here is guaranteed to be string, so that's redundant.

core/tests/tests/Number.mint
```diff
@@ -54,6 +54,17 @@ suite "Number.fromString" {
...
+    Maybe.oneOf(
+      [
+        Number.fromString(`true`),
+        Number.fromString(`0`),
+        Number.fromString(`{x: 0}`),
+        Number.fromString(`/x/`)
+      ])
+    |> Maybe.isNothing()
+  }
+
```
Sija, line 67: I don't think we ought to test the abuse of the language via backticks usage.

core/source/Number.mint
```diff
@@ -42,17 +42,21 @@ module Number {
...
       let value = Number(#{input})
 
       if (
         Number.isNaN(value)
         ||
-        Number.isNaN(value - parseFloat(#{input}))  // excludes "Infinity", "-Infinity", ""
+        #{input}.trim() === ''
```
Sija, line 54: This check should be done before constructing `value`.

Sija (changes requested):

core/source/Number.mint
nilslindemann: @gdotdesign requested it [here](https://github.com/mint-lang/mint/pull/612#discussion_r1210182549) and [here](https://github.com/mint-lang/mint/pull/612#discussion_r1210226658). You even upvoted the second comment :-)

core/tests/tests/Number.mint
nilslindemann: Yes, but as it was explicitly requested, we should also add tests for it in this case, in my opinion.

core/source/Number.mint
nilslindemann: Actually, I tested this and the version with late string trimming runs a bit faster: (check console)

```html
<script>
let x, startTime, endTime


function runParseFloat(times) {
    for (let y=0; y < times; y++) {
        x = parseFloat("1")
        x = parseFloat("1.0")
        x = parseFloat(" 1 ")
        x = parseFloat("1e-2")
        x = parseFloat("1E+2")
        x = parseFloat("1.0 xxx")
        x = parseFloat("")
        x = parseFloat("Infinity")
        x = parseFloat("-Infinity")
        x = parseFloat("x")
    }
}
startTime = performance.now()
runParseFloat(10000000)
endTime = performance.now()
console.log(`Call to runParseFloat took ${endTime - startTime} milliseconds`)



function Number_FromStringEarlyTrim(input) {
    if (typeof input !== 'string' || input.trim() === '') {
        return null
    }

    let value = Number(input)

    if (
        Number.isNaN(value)
    ) {
        return null
    }

    return value
}
function runNumber_FromStringEarlyTrim(times) {
    for (let y=0; y < times; y++) {
        x = Number_FromStringEarlyTrim("1")
        x = Number_FromStringEarlyTrim("1.0")
        x = Number_FromStringEarlyTrim(" 1 ")
        x = Number_FromStringEarlyTrim("1e-2")
        x = Number_FromStringEarlyTrim("1E+2")
        x = Number_FromStringEarlyTrim("1.0 xxx")
        x = Number_FromStringEarlyTrim("")
        x = Number_FromStringEarlyTrim("Infinity")
        x = Number_FromStringEarlyTrim("-Infinity")
        x = Number_FromStringEarlyTrim("x")
        x = Number_FromStringEarlyTrim("1x")
    }
}
startTime = performance.now()
runNumber_FromStringEarlyTrim(10000000)
endTime = performance.now()
console.log(`Call to runNumber_FromStringEarlyTrim took ${endTime - startTime} milliseconds`)




function Number_FromString(input) {
    if (typeof input !== 'string') {
        return null
    }

    let value = Number(input)

    if (
        Number.isNaN(value)
        ||
        input.trim() === ''
    ) {
        return null
    }

    return value
}
function runNumberFromString(times) {
    for (let y=0; y < times; y++) {
        x = Number_FromString("1")
        x = Number_FromString("1.0")
        x = Number_FromString(" 1 ")
        x = Number_FromString("1e-2")
        x = Number_FromString("1E+2")
        x = Number_FromString("1.0 xxx")
        x = Number_FromString("")
        x = Number_FromString("Infinity")
        x = Number_FromString("-Infinity")
        x = Number_FromString("x")
        x = Number_FromString("1x")
    }
}
startTime = performance.now()
runNumberFromString(10000000)
endTime = performance.now()
console.log(`Call to runNumberFromString took ${endTime - startTime} milliseconds`)




function runNumber(times) {
    for (let y=0; y < times; y++) {
        x = Number("1")
        x = Number("1.0")
        x = Number(" 1 ")
        x = Number("1e-2")
        x = Number("1E+2")
        x = Number("1.0 xxx")
        x = Number("")
        x = Number("Infinity")
        x = Number("-Infinity")
        x = Number("x")
    }
}
startTime = performance.now()
runNumber(10000000)
endTime = performance.now()
console.log(`Call to runNumber took ${endTime - startTime} milliseconds`)





// console.log(Number_FromString('Infinity'))
// console.log(Number_FromString(''))
// console.log(Number_FromString('   '))
// console.log(Number_FromString(true))
// console.log(Number_FromString("1"))
// console.log(Number_FromString("1.0"))
// console.log(Number_FromString(" 1 "))
// console.log(Number_FromString("1e-2"))
// console.log(Number_FromString("1E+2"))
// console.log(Number_FromString("1.0 xxx"))
// console.log(Number_FromString(""))
// console.log(Number_FromString("Infinity"))
// console.log(Number_FromString("-Infinity"))
// console.log(Number_FromString("x"))

</script>
```

core/source/Number.mint
nilslindemann: I do not get why this shows "outdated" now. Did you remove this code? Please do not.

nilslindemann:
Why does one of these reviews show "Outdated"?

gdotdesign (approved):

core/source/Number.mint
gdotdesign: I know it's redundant but I did request it, better to have it then someone creating an issue about it in the future.

core/source/Number.mint
gdotdesign: > Give me some practical use cases for Infinity, that justify this overhead for library users.

I can't really :slightly_smiling_face: but it could happen to somebody (that's why for example `toFixed` raises an error on big `digits` argument) but until it does, we can leave the things as is.

core/source/Number.mint
Sija: I'm lil' wary going down that round, since it would mean adding such guards all over the codebase - in a typed language after all.

@nilslindemann I can't find the comments you're referring to 😕

nilslindemann: I also was like, "why check for it, Mint will anyway take care of it" when I implemented it the first time. There are also performance considerations when these checks are done everywhere. I agree with Sija, in general the users should be on their own when they use backticks.

core/source/Number.mint
nilslindemann: There is actually a bug anywhere in my code which distorts the results, but I can not find it. (some hidden JavaScript optimization?) The following code gives the realistic (and expected) results:

```html
<script>

    let VALUES = ["1", "1.0", " 1 ", "1e-2", "1E+2", "1.0 xxx", "", "Infinity", "-Infinity", "x"],
        TIMES = 10000000

    function run(desc, func) {
        let startTime, endTime, x, y, values = VALUES, times = TIMES
        startTime = performance.now()
        for (y = 0; y < times; y++) {
            for (value of values) {
                x = func(value)
            }
        }
        endTime = performance.now()
        console.log(`${times} calls to ${desc} take ${endTime - startTime} milliseconds`)
    }

    run('parseFloat', parseFloat)

    run('Number', Number)

    run('NumberFromStringLateTrim', function (input) {
        if (typeof input !== 'string') {
            return null
        }
        let value = Number(input)
        if (
            Number.isNaN(value)
            ||
            input.trim() === ''
        ) {
            return null
        }
        return value
    })

    run('NumberFromString', function (input) {
        if (typeof input !== 'string' || input.trim() === '') {
            return null
        }
        let value = Number(input)
        if (
            Number.isNaN(value)
        ) {
            return null
        }
        return value
    })

    run('NumberFromStringNoTypecheck', function (input) {
        let value = Number(input)
        if (
            Number.isNaN(value)
            ||
            input.trim() === ''
        ) {
            return null
        }
        return value
    })

    // run('', function (input) {
    // })

</script>
```

In my browser this results in

```
10000000 calls to parseFloat take 2594.7000000001863 milliseconds
10000000 calls to Number take 2872.5 milliseconds
10000000 calls to NumberFromStringEarlyTrim take 3619.5 milliseconds
10000000 calls to NumberFromString take 3617.199999999255 milliseconds
10000000 calls to NumberFromStringNoTypecheck take 3541.600000000559 milliseconds
10000000 calls to Angular Version take 3245.5999999996275 milliseconds
```

We can see here

* As expected, the native functions run fastest (and `parseFloat` is actually the fastest).
* It is actually not relevant if the trimming happens early or late, so I will change the commit to do it early, like @Sija suggested. It is more readable indeed.
* The type checking is not expensive.
* The Angular Version (original commit) is slightly faster than the other non-native functions

My assumption that the code runs faster now was wrong, it actually runs a bit slower. Considerable, if you let it run ten million times, *looks worried*, haha, no.

nilslindemann:
I do not know why the `Window.href` test fails in the macos-latest test.

Sija:
@nilslindemann Don't worry about it, it's a flaky spec (we ought to do sth about finally 😅)

Sija:
@nilslindemann Don't worry about it, it's a flaky spec (we ought to do sth about finally 😅)

core/source/Number.mint
Sija: @gdotdesign What's the reason for having such guard here?

gdotdesign: Just backticks :joy: I can let it go.

Sija: Hahaha, so let it go plz, since if this passes I'm gonna question the type-safety of the language :P

nilslindemann:
The testing solves problems and creates new ones :-)

gdotdesign (approved):

Sija (approved):

nilslindemann:
Well, that was a difficult birth :-)

nilslindemann:
macOS is now also nice again.

PR #618, opened 2023-06-26
Localization Language Feature 

gdotdesign:
This PR adds language features for handling localizations: https://github.com/mint-lang/mint/blob/locales/documentation/Language/Locale.md


documentation/Language/Locale.md
```diff
@@ -0,0 +1,47 @@
...
+  }
+}
+```
+
+A locale consists of a tree structure where the keys are lowercase identifiers and the values are expressions.
+
+To localize a value you need to use the locale token:
+
+```
+:ui.buttons.ok
```
Sija, line 28: Using an `:atom` token seems IMO too generic and kinda unnecessary, I'd rather go with regular function in the global scope, sth like `t(...)`.

documentation/Language/Locale.md
```diff
@@ -0,0 +1,47 @@
...
+This feature of the language allows specifing localization tokens and values for languages indentified by [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) codes.
+
+```mint
+locale en {
+  ui: {
+    buttons: {
+      ok: "OK"
+    }
+  }
+}
```
Sija, line 12: It's a good start! I like the explicit locale strings definitions 👍 
How do you express the variables/interpolation within strings?
Also, this doesn't seem to support explicit pluralization rules.

gdotdesign: It's not just strings, the values can be any type / expression so it's easy to do interpolation / pluralization with functions:

```mint
locale en {
  greet: (name : String) { "Hello #{name}!" },
  article: (howMany : Number) {
    if (howMany == 0 && howMany == 1) {
      "Article"
    } else {
      "Articles"
    }
  }
}
```
The type signature of the values must match between locales.

documentation/Language/Locale.md
gdotdesign: There are no global functions in Mint and they need to learn that as well, also if we would use that it could conflict with any locally defined function with the same name.

We could put it on `Locale` module / store (`Locale.t`) but it would make it harder to connect the components to it in order to trigger a re-render when the locale changes https://github.com/mint-lang/mint/pull/618/files#diff-2a98a3d17ef6693f5e57ca9b02ea16a62585b9b2e038e525c9ce7ae0cb4c3c91R146 and also it would include a lot of boilerplate.

Localization is doable without a language feature, but with a lot of boilerplate:
```
// Need to define the type and subtyles for the locale
record Locale {
  navigation : Locale.Navigation,
  queryString : Locale.Language,
  ...
}

record Locale.Navigation {
  ...
}

record Locale.Language {
  ...
}

// Need to define locales in a module
module Locales {
  const EN =
    {
      queryString: ...
      navigation: ...
    }

// Need to have a store to house the current locale
store Locale {
  state locale : Locale = Locales.EN

  fun setLocale (string : String) {
    ...
  }

  const LOCALES =
    [
      {"EN", locale.languages.en},
      {"HU", locale.languages.hu},
      {"ES", locale.languages.es}
    ]
    |> Map.fromArray
}

// Connect to that store
component X {
  connect Locale exposing { locale }

  fun render {
    locale.queryString.someKey
  }
}
```

So this feature is basically the same but without the boilerplate.

documentation/Language/Locale.md
Sija: I see. It's an option, alright, although it seems like a PITA to repeat a lot of boilerplate code in order to achieve these two functionalities.

gdotdesign: What boilerplate code are you referring to? For each language?

jansul: I appreciate that the values are just expressions. I've not done much localisation but I assume it can get quite complicated depending on the target language.

Perhaps for some common tasks there could be some methods in the core library?

```mint
locale en {
  article: Localisation.plural("No Articles", "1 Article", (count : String) { "#{count} Articles" })
}
```

Sija: @gdotdesign I meant pluralization, @jansul beat me to it :)

gdotdesign: There is no reason why we can't have functions like in the `Locale` module but I'm not sure that all languages have pluralization like that :thinking: which a quick google search confirms https://stackoverflow.com/questions/6166064/i18n-pluralization :sweat_smile: 

So maybe we will need to have separate modules for each languages functions `Locale.EN.pluralize()`, `Locale.RU.pluralize()`, etc...

jansul: Separate module per language seems to make sense, especially given that Russian example!

My only other concern for this feature would be the file size if an application includes a lot of languages, but that would have to wait until something like code-splitting was added to Mint.

Sija: I'd try to generalize it within a single module, but load the language rules dynamically. Having `zero`, `one`, `few`, `many` and `other` keys (with fallbacks if undefined, e.g. `few || other`) should handle most of the cases. Kinda like how Rails is [handling](https://guides.rubyonrails.org/i18n.html#pluralization) that.

farism:
A couple of localization thoughts just from experience with it in our prod app:

1. There can potentially be hundreds/thousands of lines of text in a large application, multiplied by 10+ languages can easily add up to hundreds of kilobytes in your bundle. It would be nice from an optimization standpoint if each there was an option to have each locale code split out and loaded on demand.
2. When using a localization service ([Phrase](https://phrase.com/), [Smartling](https://www.smartling.com/), and other services like this) they generally expect to be working with some standardized data exchange format like `.yaml` or `.json`. Something to think about from a workflow perspective.
3. Packages may need to ship with their own translations, and whatever system Mint goes with should have seamless integration between the package and the app. I think having the `locale` scope be open solves this issue? But not entirely sure if there are some other nuances.

documentation/Language/Locale.md
gdotdesign: I think it's out of the scope of the current PR :slightly_smiling_face: so I'm resolving this in order to get it merge.

Sija (commented):
I like the approach of making the translations a first class citizen. I believe though that the implementation should handle the pluralization, loading data from files and the other points raised by @farism.

gdotdesign:
> I like the approach of making the translations a first class citizen. I believe though that the implementation should handle the pluralization, loading data from files and the other points raised by @farism.

I agree on all parts, but it needs to be a multi-stage process, we can't implement everything in one PR.

1. I'm currently refactoring the compiler to allow dynamic imports, which will include locales.
2. This is the hardest part to implement, since I presume everyone has its own format, so we either need to decide which format to use or we keep it as is and build the tooling around importing / exporting formats.
3. The locale definitions are open, and if packages use namespaces it shouldn't be a big issue. The only thing we might need is to allow overriding a specific key but I wouldn't implement this until we run into the need in production.
4. Pluralization and the whole translation logic is so wide (date-time, currencies, etc...) is that I fear that having them as language features are not really feasible.

src/ast.cr
```diff
@@ -111,6 +115,18 @@ module Mint
             )
           end
 
+      @unified_locales =
+        @locales
+          .group_by(&.language)
+          .map do |_, locales|
+            Locale.new(
+              input: Data.new(input: "", file: ""),
+              fields: locales.flat_map(&.fields),
```
jansul, line 124: I'm guessing whichever `locale` is parsed last is the one that's used? (for the same language)

core/source/Locale.mint
```diff
@@ -0,0 +1,19 @@
+module Locale {
+  /* Sets the current locale. */
+  fun set (locale : String) : Bool {
```
jansul, line 3: May be worth making the response to this `Promise(Bool)` in anticipation of loading in languages asynchronously?

documentation/Language/Locale.md
```diff
@@ -0,0 +1,49 @@
...
+To localize a value you need to use the locale token:
+
+```
+:ui.buttons.ok
+```
+
+or if it's a function if can be called:
+
+```
+:ui.buttons.ok(param1, param2)
```
jansul, line 34:
```suggestion
locale en {
  ui: {
    buttons: {
      ok: (param1 : String, param2 : String) { 
        "Button #{param1} #{param2}!" 
      }
    }
  }
}

:ui.buttons.ok(param1, param2)
```


May be worth adding an example of a function being used in the documentation for the visual learners :smile:

jansul (approved):
I would generally agree with @gdotdesign, there doesn't seem to be anything about this design that would prevent implementing most of the features mentioned. 

The only awkward thing might be:

> 2\. When using a localization service ([Phrase](https://phrase.com/), [Smartling](https://www.smartling.com/), and other services like this) they generally expect to be working with some standardized data exchange format like `.yaml` or `.json`. Something to think about from a workflow perspective.

However I don't think it's worth giving up the flexibility of the values being expressions to support these services. Hopefully we could find a nice way of importing/exporting locales in a way that works with them.

core/source/Locale.mint
gdotdesign: Yes, good idea, will update it :+1:

src/ast.cr
gdotdesign: No, it's merging all the keys from all the instances.

PR #615, opened 2023-06-13
Semantic Tokenizer

gdotdesign:
This PR adds support for [semantic tokenizer](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens) for the language server and also for other uses.

Features:

- [x] Language Server - syntax highlighting for code editors
- [x] `highlight` command - syntax highlight specific files	in the terminal or as HTML files
- [x] `@highlight` language feature - same as `@format` but also highlights the given code as HTML

Fixes:

- [x] Parsed nodes are kept in the `ast.nodes` array when parsing a parent node fails.

Future work:

This can be used to provide syntax highlighted error messages :tada:

PR #619, opened 2023-07-05
LSP: Add Go to Definition support for `Ast::Variable`

jansul:
Been through a couple of iterations of this and still not terribly happy with the code (could do with another pass or two :sweat_smile:), but this PR adds "Go to Definition" support for the majority of `Ast::Variable`'s.

Here's a quick example of it in action:

https://github.com/mint-lang/mint/assets/1927518/421bb271-328d-43ab-874b-3e2709eec1fd

Will keep testing as I'm sure there's a bunch of weird and wonderful ways of using Mint that I haven't accounted for, but it's pretty handy as is!

_Note_: There are a couple of tests that will fail until the fix from https://github.com/mint-lang/mint/pull/615 has been merged

### DIFF
--- a/src/ls/definition/enum_id.cr
+++ b/src/ls/definition/enum_id.cr
@@ -8,7 +8,11 @@ module Mint
         if name.nil?
           stack.each do |parent|
             case parent
-            when Ast::Component, Ast::Store
+            when Ast::Component,
+                 Ast::Store,
+                 Ast::Suite,
+                 Ast::Module,
+                 Ast::Provider
               parent.constants.each do |constant|
                 if node.option.value == constant.name.value
                   return location_link server, node.option, constant.name, constant
--- a/src/ls/definition/enum_id.cr
+++ b/src/ls/definition/enum_id.cr
@@ -2,20 +2,35 @@ module Mint
   module LS
     class Definition < LSP::RequestMessage
       def definition(node : Ast::EnumId, server : Server, workspace : Workspace, stack : Array(Ast::Node))
-        return unless name = node.name
-        return unless enum_node =
-                        workspace.ast.enums.find(&.name.value.==(name.value))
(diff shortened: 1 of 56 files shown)


src/ls/definition/enum_id.cr
```diff
@@ -2,20 +2,35 @@ module Mint
...
-        return unless enum_node =
-                        workspace.ast.enums.find(&.name.value.==(name.value))
+        name = node.name
 
-        return if Core.ast.enums.includes?(enum_node)
+        # When `.name` is nil the node is used as a CONSTANT
+        if name.nil?
+          stack.each do |parent|
+            case parent
+            when Ast::Component, Ast::Store
```
gdotdesign, line 11: It can also be `Ast::Suite`, `Ast::Module` and `Ast::Provider`.

gdotdesign (changes requested):

gdotdesign:
It looks very good! Thank you for adding a huge amount of tests :pray: Please update the PR from the master branch and it should be ready to go from my side :+1:

Sija (approved):
LGTM

src/ls/definition/enum_id.cr
gdotdesign: > I had a quick look and it seems they aren't being formatted (so are removed when you format the file) and also not type checked properly. Can look into fixing that in a separate PR?

Nice find :+1: yeah go ahead, should be a quick fix.

jansul: Updated to support `Ast::Provider` now too :+1: 

Haven't had much chance to test this PR in earnest yet (unfortunately have to use Typescript in work :sob:) but overall should work OK. Let me know if you find any issues!

gdotdesign (approved):

PR #621, opened 2023-07-22
Fix constants/gets not working within a Provider

jansul:
As mentioned within https://github.com/mint-lang/mint/pull/619#discussion_r1270890982, this PR fixes a couple of issues with both constants and gets when using them with a Provider:

* Constants were being removed when formatting
* Constants weren't being type checked correctly (you would receive a `VariableMissing` error)
* Gets were not being parsed 

I've done a quick test in browser and the above seems to be working there too :smile: Unfortunately none of the existing providers use these features yet - perhaps some `.mint` tests purely for testing language features rather than the standard library might be useful? 

I'll update the other PR to support Providers once this is merged :+1:

### DIFF
--- a/spec/compilers/provider_with_items
+++ b/spec/compilers/provider_with_items
@@ -1,0 +1,102 @@
+record Subscription {
+  a : Bool,
+  b : Bool
+}
+
+provider Provider : Subscription {
+  const NAME = "hello"
+
+  state a : String = ""
+
+  get b : String {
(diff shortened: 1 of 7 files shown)


Sija (approved):

gdotdesign (approved):

gdotdesign:
Nice, thank you very much :pray:

PR #623, opened 2023-08-05
LSP: Add Go To Definition support for `Ast::ModuleAccess`

jansul:
Fairly straightforward one, this adds Go To Definition support for `Ast::ModuleAccess` as demonstrated in the video below:

https://github.com/mint-lang/mint/assets/1927518/b76a52e9-acae-40d9-80a2-4549b9e20d90

While allowed by the language, some of these examples will error when running them [in the browser](https://sandbox.mint-lang.com/sandboxes/lN4-HdUp6_JfXw). If you like I can have a look at why next if you like?

As part of this PR, I have removed the file `src/parsers/constant_variable.cr` and replaced it with the new method for parsing constants introduced in https://github.com/mint-lang/mint/pull/594. This old version still had the bug which allowed invalid constants.


Sija (approved):
That's rad! Love the test coverage ❤️

jansul:
OK, turns out it was less straightforward than I thought :sweat_smile: 

I forgot that you can:
 * Have multiple modules with the same name
 * Modules can share their names with enums/records

This PR should now work correctly with both. For the former, it will now link to all instances of a module. Here's how that looks like in VSCode:

https://github.com/mint-lang/mint/assets/1927518/a9db14c8-ec42-45bc-bcf0-1f28c4861cac

src/ls/definition.cr
```diff
@@ -4,7 +4,7 @@ module Mint
     class Definition < LSP::RequestMessage
       property params : LSP::TextDocumentPositionParams
 
-      def execute(server) : LSP::LocationLink | LSP::Location?
+      def execute(server) : Array(LSP::LocationLink | LSP::Location) | LSP::LocationLink | LSP::Location | Nil
```
Sija, line 7: I'd prefer a more unified return type `Array(LSP::LocationLink | LSP::Location)`

Sija (changes requested):

src/ls/definition.cr
jansul: I thought the above matched the [specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_typeDefinition), but it turned out it doesn’t!

The return type for “Go to definition” can be any of `Location | Location[] | LocationLink[] | null`, but a singular `LocationLink` is not allowed - so we are currently out of spec 😅

Simplifying it to `Array(LSP::LocationLink | LSP::Location)` might be a good option, will have to have a think about how best to refactor all the definition methods.

gdotdesign: > Simplifying it to `Array(LSP::LocationLink | LSP::Location)` might be a good option, will have to have a think about how best to refactor all the definition methods.

IMO, that can be in a separate PR.

gdotdesign (approved):

Sija (approved):

PR #624, opened 2023-08-08
Separate nightly CI runs into a separate workflow.

gdotdesign:
For a few months now, the CI task was/is broken because the builds fail on the nightly builds. Mostly because of ameba and Crystal and it doesn't seem to be getting better (I was hoping for 1.9 to fix this but it just got broken again).

This causes some issues:
- the badge in the README is red, signaling that the project is neglected
- perfectly fine pull requests appear to be broken
- it takes time to track down the issue and get it fixed for a Crystal version that is not released yet 
- I get emails for all failures

So this PR separates the Crystal nightly builds into it's on workflow, which runs on every Monday.

### DIFF
--- a/.github/workflows/ci-nightly.yml
+++ b/.github/workflows/ci-nightly.yml
@@ -1,12 +1,8 @@
-name: CI
+name: CI Nightly
 
 on:
-  push:
-    branches: [master]
-  pull_request:
-    types: [opened, synchronize, reopened]
-  release:
-    types: [published]
+  schedule:
+    - cron: "0 6 * * 1"
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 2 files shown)


jansul:
May be worth adding Windows to the nightly one? Will still fail for a good while; there's a few things still needed outside Mint for it to actually work, but would be nice to see that green tick come through eventually!

Sija (approved):
LGTM 💯

gdotdesign:
> May be worth adding Windows to the nightly one? Will still fail for a good while; there's a few things still needed outside Mint for it to actually work, but would be nice to see that green tick come through eventually!

That's a good point, I'll try to do it in a separate PR :slightly_smiling_face:

PR #625, opened 2023-08-08
Named Arguments

gdotdesign:
This PR implements the ability to call functions with an alternative syntax using the names of the arguments.

```
component Main {
  fun test (argument1 : String, argument2: Number) : Html {
    <div/>
  }

  fun render : Html {
    test(argument2: 0, argument1: "")
  }
}
```

The only rule as of now is that there can't be normal arguments and labelled arguments in a call.


src/type_checker/type.cr
```diff
@@ -18,7 +19,16 @@ module Mint
...
+            parameters.map do |param|
+              pretty =
+                param.to_pretty.as(String)
+
+              if param.label
+                "#{param.label}: #{pretty}"
+              else
+                pretty
+              end
+            end
```
Sija, line 31: I'd refactor this into `Record#to_pretty(named_arguments: true)`.

src/type_checkers/argument.cr
```diff
@@ -3,7 +3,9 @@ module Mint
     def check(node : Ast::Argument) : Checkable
       node.default.try { |default| resolve default }
 
-      resolve_type(resolve(node.type))
+      resolve_type(resolve(node.type)).tap do |item|
+        item.label = node.name.try(&.value)
+      end
```
Sija, line 8:
```suggestion
      resolve_type(resolve(node.type))
        .tap(&.label = node.name.try(&.value))
```

src/type_checkers/call.cr
```diff
@@ -2,9 +2,17 @@ module Mint
...
     type_error CallArgumentTypeMismatch
+    type_error CallNotFoundArgument
+    type_error CallMixedArguments
     type_error CallTypeMismatch
     type_error CallNotAFunction
 
+    def check(node : Ast::CallExpression)
+      resolve(node.expression).dup.tap do |item|
+        item.label = node.name.try(&.value)
+      end
```
Sija, line 13:
```suggestion
      resolve(node.expression)
        .dup
        .tap(&.label = node.name.try(&.value))
```

src/type_checkers/call.cr
```diff
@@ -38,7 +46,33 @@ module Mint
       } if node.arguments.size > argument_size ||       # If it's more than the maximum
            node.arguments.size < required_argument_size # If it's less then the minimum
 
-      node.arguments.each_with_index do |argument, index|
+      args =
+        if node.arguments.all?(&.name.nil?)
+          node.arguments
+        elsif node.arguments.all?(&.name.!=(nil))
```
Sija, line 52:
```suggestion
        case node.arguments
        when .all?(&.name.nil?)
          node.arguments
        when .none?(&.name.nil?)
```

Sija (approved):
Nice 🤩

Sija:
As a side note: I'd use _Calls with named arguments_ or _Named arguments in calls_ terms rather than _Labeled calls_, which IMO is a bit abstract.

PR #627, opened 2023-08-17
Refactoring Errors

gdotdesign:
This PR refactors errors. Since it's a big PR, I've opted to do each entity in a separate commit (for parser and type checker) where tests pass for each commit so not to really break anything.

The main change is different errors are removed `TypeError`, `SyntaxError`, etc... and replaced with just one error `Error` separate error files are moved into their call site. The message builder pattern has been removed and it's in the `Error` constructor now.

* parser specs and type checker specs are merged into one specs (`spec/examples/*`)
* macros for errors are removed
* the syntax of error messages have been simplified

Probably the best commit to check is this one https://github.com/mint-lang/mint/pull/627/commits/d1e7b82b0b13ae3362b36f09347d3c74e5dfd611 since it introduces the new error system and this one https://github.com/mint-lang/mint/pull/627/commits/5c66f3bed515f5931c3fbe2288cb518a998f398d which removes the old error code.

So in short, errors are easier to reason about and have less magic in them :slightly_smiling_face: 

This will be in the 0.20.0 release. I'll go over all the error messages to check them in another PR.


spec/language_server/definition/location_link/variable_store_get
```diff
@@ -61,8 +61,8 @@ store Test {
         "character": 2
       },
       "end": {
-        "line": 5,
-        "character": 2
+        "line": 3,
+        "character": 3
```
jansul, line 65: Nice removing the additional whitespace from these nodes! :smile:

jansul:
Always impressed by how much you are able to achieve in your spare time! :see_no_evil: 

Looks great, consolidation of the errors and specs does seem like it simplifies things a fair bit. :raised_hands: 

I assume one of the goals for this was to allow for more expressive errors in the future? (as you are less limited by the error DSL)

jansul:
Have been playing around with implementing the [textDocument/publishDiagnostics](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_publishDiagnostics) notification message. It's all purely bodged in at the moment :sweat_smile: but here's how it currently looks in VSCode:

https://github.com/mint-lang/mint/assets/1927518/5eeb2567-de2e-4a86-841a-70e1f5948465

And here's how it looks in Sublime:

https://github.com/mint-lang/mint/assets/1927518/cfc9fdce-75d3-41ef-b34f-190bc3e624e0

There are a few things that make this LS feature somewhat hard to implement at the moment. So I thought it might be worth highlighting them as it could influence how you continue with this refactoring. Apologies in advance for the wall of text :see_no_evil: 

The problems at the moment are:

* Each [Diagnostic](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic) sent in the message requires a [Range](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#range) to be included along with it. This is what is used to draw the "red squiggly lines" (in case of an error diagnostic)

  Currently there isn't an easy way of determining what that range could be for all errors. For the above demonstration I just search through the [locals](https://github.com/mint-lang/mint/blob/master/src/errors/error.cr#L11) of an error for an `Ast::Node` but that isn't always correct.

  Perhaps that the range (or something to get the range from, like a node) could be passed along with each error?

  ```crystal
  error :my_custom_error, range | node do
  end
  ```
* Rendering the error using the terminal renderer works surprisingly OK - especially in Sublime Text! The only place it doesn't look great is in the "Problems" tab within VSCode, where the code snippet isn't rendered using a mono-space font.
  
  It seems most language servers limit the diagnostic message to at most a couple of lines (with the first line usually being pretty descriptive)

  We could probably also have single line versions of the error message just using a big switch:

  ```crystal
   case error_name
   when :my_first_error  then "My first error"
   when :my_second_error then "My second error"
   ....
   end
   ```

  Without the rest of the context of the error though, we might be limited in how expressive we can be with the messages.

  However as both Sublime and VSCode support showing both the hover and error message within in the same panel, we could get the best of both worlds by combing both:

  image

gdotdesign:
I'm closing this since I merged a bigger refactor into `master` which includes these changes. I'll try to integrate the range information for the comment above from @jansul

PR #628, opened 2023-08-19
LSP: Fix enum destructuring/id linking to other types with the same name

jansul:
Noticed a bug when working on a fix for the [issue mentioned in my last PR](https://github.com/mint-lang/mint/pull/623#discussion_r1285119505).

When doing "Go to Definition" on an `Ast::EnumId` or `Ast::EnumDestructuring` node, if their `.option` field matched the name of an existing component, store etc, then it would link to that instead.

Here's an example of the issue in video form:

https://github.com/mint-lang/mint/assets/1927518/25c6b32c-c312-43b9-82da-8d5799af3db8

The fix here was to stop tracking `Ast::TypeId` within these two nodes, as hover/definition etc use the higher level nodes to provide their features anyway.

### DIFF
--- a/spec/language_server/definition/location_link/enum_destructuring_option
+++ b/spec/language_server/definition/location_link/enum_destructuring_option
@@ -1,3 +1,9 @@
+component Error {
+  fun render {
+    <div/>
+  }
+}
+-------------------------------------------------------------file component.mint
 /* Comment for Status enum. */
 enum Status {
   Error
--- a/spec/language_server/definition/location_link/enum_id_option
+++ b/spec/language_server/definition/location_link/enum_id_option
@@ -1,3 +1,9 @@
+component Ok {
+  fun render {
+    <div/>
+  }
+}
+-------------------------------------------------------------file component.mint
 enum Status {
   Error
   Ok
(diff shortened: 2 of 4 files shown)


gdotdesign:
> The fix here was to stop tracking `Ast::TypeId` within these two nodes, as hover/definition etc use the higher level nodes to provide their features anyway.

Would this mean that the options are not linkable themselves, just the enum itself?

jansul:
> Would this mean that the options are not linkable themselves, just the enum itself?

That still works thanks to the use of `cursor_intersects?` within the [enum_id definition file](https://github.com/mint-lang/mint/blob/master/src/ls/definition/enum_id.cr#L32). 

In hindsight we might not actually need that method, instead just looking through the stack as we do [for module access.](https://github.com/mint-lang/mint/blob/master/src/ls/definition/type_id.cr#L5). But possible one for another refactoring 🙈

gdotdesign:
> That still works thanks to the use of `cursor_intersects?` within the [enum_id definition file](https://github.com/mint-lang/mint/blob/master/src/ls/definition/enum_id.cr?rgh-link-date=2023-08-21T10%3A42%3A31Z#L32).

Ah yeah, that was what I was missing :)

PR #630, opened 2023-08-20
Fix keywords and operators being kept when parsing a parent node fails

jansul:
Working on refactoring how `Workspace`'s work within the language server which will require updating [mint-lang/mint-vscode](https://github.com/mint-lang/mint-vscode). 

Thought while I was there I would test out [semantic tokens](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens) and noticed a couple of bugs. The below is being highlighted using the tokens from language server only: 

Screenshot from 2023-08-20 22-17-15

The issues are: 

* Multiple line comments aren't highlighted correctly. This is because VSCode doesn't support the `multilineTokenSupport` feature (other editors can probably handle it fine!)
* The `next` part of `nextTitle` is being highlighted as if it was a keyword.

This PR fixes the latter. The issue is very similar to the one that was fixed as part of the initial PR https://github.com/mint-lang/mint/pull/615 however this one involves keywords and operators instead of nodes!

I solved it in a very dumb way, just copying and pasting the existing logic. :sweat_smile: Only `keyword`'s cause this specific bug, but I believe it may also be possible for `operator`'s to cause similar bugs so I included that too. 

Here's how it looks after the fix:

Screenshot from 2023-08-20 22-36-42

### DIFF
--- a/src/parser.cr
+++ b/src/parser.cr
@@ -22,16 +22,27 @@ module Mint
 
     def start(&)
       start_position = position
-      node_size = ast.nodes.size
+
+      nodes_size = ast.nodes.size
+      keywords_size = ast.keywords.size
+      operators_size = ast.operators.size
 
       begin
         node = yield position
         @position = start_position unless node
-        ast.nodes.delete_at(node_size...) unless node
+
+        ast.nodes.delete_at(nodes_size...) unless node
+        ast.keywords.delete_at(keywords_size...) unless node
+        ast.operators.delete_at(operators_size...) unless node
+
         node
       rescue error : Error
         @position = start_position
-        ast.nodes.delete_at(node_size...)
+
+        ast.nodes.delete_at(nodes_size...)
+        ast.keywords.delete_at(keywords_size...)
+        ast.operators.delete_at(operators_size...)
+
         raise error
       end
     end
--- a/src/parser.cr
+++ b/src/parser.cr
@@ -31,9 +31,11 @@ module Mint
         node = yield position
         @position = start_position unless node
 
-        ast.nodes.delete_at(nodes_size...) unless node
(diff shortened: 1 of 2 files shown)


src/parser.cr
```diff
@@ -22,16 +22,27 @@ module Mint
...
+      operators_size = ast.operators.size
 
       begin
         node = yield position
         @position = start_position unless node
-        ast.nodes.delete_at(node_size...) unless node
+
+        ast.nodes.delete_at(nodes_size...) unless node
+        ast.keywords.delete_at(keywords_size...) unless node
+        ast.operators.delete_at(operators_size...) unless node
```
Sija, line 36:
```suggestion
        unless node
          ast.nodes.delete_at(nodes_size...)
          ast.keywords.delete_at(keywords_size...)
          ast.operators.delete_at(operators_size...)
        end
```

gdotdesign:
> * Multiple line comments aren't highlighted correctly. This is because VSCode doesn't support the `multilineTokenSupport` feature (other editors can probably handle it fine!)

I didn't know that, I use Sublime Text and it does support it. Nonetheless, we should support honor the flag properly.

> I solved it in a very dumb way, just copying and pasting the existing logic. 😅 Only `keyword`'s cause this specific bug, but I believe it may also be possible for `operator`'s to cause similar bugs so I included that too.

I like simple :wink: 

With @Sija comment fixed, this should be ready to merge :tada:

jansul:
> I didn't know that, I use Sublime Text and it does support it. Nonetheless, we should support honor the flag properly.

Agreed 👍 Shouldn't be too tough to support it I think. Was surprised how well it works with only semantic tokens supplying the highlighting though!

Sija (approved):

PR #631, opened 2023-08-26
Add missing Math constants and functions

farism:
Tries to achieve 1:1 parity with the browser `Math` APIs

### DIFF
--- a/core/source/Math.mint
+++ b/core/source/Math.mint
@@ -1,12 +1,27 @@
 /* Mathematical functions. */
 module Math {
+  /* Euler's number and the base of natural logarithms; approximately `2.718`. */
   const E = `Math.E`
+
+  /* Natural logarithm of `10`; approximately `2.303`. */
   const LN10 = `Math.LN10`
+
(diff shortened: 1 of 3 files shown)


core/source/Math.mint
```diff
@@ -1,5 +1,14 @@
 /* Mathematical functions. */
 module Math {
+  const E = `Math.E`
+  const LN10 = `Math.LN10`
+  const LN2 = `Math.LN2`
+  const LOG10E = `Math.LOG10E`
+  const LOG2E = `Math.LOG2E`
+  const PI = `Math.PI`
+  const SQRT12 = `Math.SQRT1_2`
+  const SQRT2 = `Math.SQRT2`
```
gdotdesign, line 10: Would be nice if these constants have comments as well.

gdotdesign (approved):

core/source/Math.mint
farism: Added 👍

Sija (approved):

PR #653, opened 2023-09-12
Removed deprecated 'with' from project init scaffold and small bug fix

jrrom:
1. Removed deprecated `with` from project init scaffold that throws errors in Mint 0.17.0 and above.
2. Added 'firefox-bin' to `BROWSER_PATHS`. This is because in Gentoo the binary distributions of Firefox are called '[firefox-bin](https://packages.gentoo.org/packages/www-client/firefox-bin)' to differentiate it from the compiled '[firefox](https://packages.gentoo.org/packages/www-client/firefox)'.

### DIFF
--- a/src/app_init/app/tests/Main.mint
+++ b/src/app_init/app/tests/Main.mint
@@ -1,9 +1,7 @@
 suite "Main" {
   test "Greets Mint" {
-    with Test.Html {
-      <Main/>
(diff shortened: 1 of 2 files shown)


gdotdesign (approved):

gdotdesign:
Thank you! 🙏

PR #659, opened 2023-09-13
Update test runner to serve assets

farism:
I was trying to write some tests that relied on assets being served, and I noticed that they weren't being served in the test runner.

I'm not sure if there was a specific reason assets were not being served, or just an oversight.


gdotdesign:
Thanks :pray:

PR #660, opened 2023-09-13
Added bindings to the Console API.

jrrom:
Mint bindings for the Console API. It cannot be tested automatically however I ran everything and it all works.

### DIFF
--- a/core/source/Console.mint
+++ b/core/source/Console.mint
@@ -248,7 +248,7 @@ module Console {
   */
   fun timestamp (label : String = "Default") : String {
     if `!console.timestamp` {
-      Debug.log("Your browser does not support console.profile")
+      Debug.log("Your browser does not support console.timestamp")
     } else {
       `console.timestamp(#{label})`
     }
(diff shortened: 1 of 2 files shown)


core/source/Console.mint
```diff
@@ -0,0 +1,282 @@
...
+
+  /*
+  Logs the number of times that this particular call to count() has been called.
+  Returns the label passed along with the current count.
+
+  [count()](https://developer.mozilla.org/en-US/docs/Web/API/console/count)
+  */
+  fun count (label : String = "Default") : Tuple(String, Number) {
+    `console.count(#{label})`
+    Console.Counter.increment()
```
gdotdesign, line 43: The counter is global, not based on the label, I think that wasn't the intention :slightly_smiling_face: 

You need to change the state in the store to accommodate this:
```
state counts : Map(String, Number) = Map.empty()
```

gdotdesign (changes requested):

jrrom:
All done! Made requested changes and added a new test suite. That count function gave me so much misery 😢

PR #663, opened 2023-09-16
Combine `Provider.Keydown` and `Provider.Keyup` into `Provider.Keyboard`

farism:
We spoke about this briefly on Discord, and I was going to write up an issue but figured I may as well just contribute the change since it was relatively easy.

Background info:

It would be nice to have keyboard providers be more consistent with other providers, particularly `Provider.Mouse`. At the moment, keyboard events are split into two providers, one for `downs` and one for `ups`. 

This PR combines the providers into a single Provider, called `Provider.Keyboard` which has both `downs` and `ups`, which is a bit more predictable.

I also added some test utilities to `Test.Html` module in order to properly test these changes.

1. `Test.Html.triggerKeyDown(context : Context, selector : String, key : String)`
2. `Test.Html.triggerKeyUp(context : Context, selector : String, key : String)`


gdotdesign (approved):

gdotdesign:
Thanks :pray:

PR #669, opened 2023-09-18
Allow asyncronhous tests.

### LINKED ISSUE #662: Support async in tests
The following does not work

```mint
suite "async" {
  test "it can be used" {
    await Timer.timeout(1000)
    true
  }
}
```

Currently errors with

```mint
░ TYPE ERROR ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░

The type of a test does not match any of the allowed types. 

I was expecting one of: 

  Bool, Test.Context(a)

Instead it is: 

  Promise(Bool)
```

Because trying to use `await` anywhere inside of the body of a block converts the return to a `Promise`, which the test runner does not expect.

Supporting async in tests would be a great feature and definitely unlock a lot more testing capabilities.

gdotdesign:
This was pretty simple to add. Fixes #662


core/tests/tests/Test.mint
```diff
@@ -0,0 +1,11 @@
...
+  test "true" {
+    await true
+  }
+}
+
+suite "Test (Async)" {
+  test "true" {
+    await Test.Context.of(true)
+  }
+}
```
Sija, line 11:
```suggestion
suite "Test (Async)" {
  test "await Bool" {
    await true
  }

  test "await Test.Context(Bool)" {
    await Test.Context.of(true)
  }
}
```

Sija (approved):
Looks good! I've left one small suggestion.

PR #670, opened 2023-09-18
Improve `format` command.

### LINKED ISSUE #640: `mint format` issues
Running `mint format` today will not work if there are type errors anywhere in the project.

`mint format` should

- Be able to run independently from the typechecker.
- Only care about syntax. If a file can be parsed it can be formatted.
- Only care about one file at a time. Syntax errors elsewhere in the project should not prevent an individual file being formatted.

gdotdesign:
Improvements:

- Doesn't error out if there is no `mint.json` file
- Pattern has precedence over `mint.json` fields
- Error case is now not necessary probably

Fixes #640

### DIFF
--- a/src/commands/format.cr
+++ b/src/commands/format.cr
@@ -47,34 +47,41 @@ module Mint
 
       private def format_files
         current =
-          MintJson.parse_current
+          MintJson.parse_current?
 
-        format_directories =
-          current.source_directories | current.test_directories
+        files =
+          if pattern_argument = arguments.pattern.presence
+            Dir.glob(pattern_argument)
+          elsif current
+            format_directories =
+              current.source_directories | current.test_directories
 
-        format_directories_patterns =
-          format_directories.map do |dir|
-            SourceFiles.glob_pattern(dir)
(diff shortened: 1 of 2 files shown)


src/commands/format.cr
```diff
@@ -47,34 +47,41 @@ module Mint
...
-
-        if files.empty?
+        if files.try(&.empty?)
           terminal.puts "Nothing to format!"
           true
         else
           all_formatted = true
 
-          files.each do |file|
+          files.not_nil!.each do |file|
```
Sija, line 72: `files` could be a `nil` here actually 😅

src/commands/format.cr
```diff
@@ -47,34 +47,41 @@ module Mint
...
+
+        files =
+          if pattern_argument = arguments.pattern.presence
+            Dir.glob(pattern_argument)
+          elsif current
+            format_directories =
+              current.source_directories | current.test_directories
+
+            format_directories_patterns =
+              format_directories.map do |dir|
```
Sija, line 60: This could be a `map!` call, I think.

PR #672, opened 2023-09-19
Allow functions to be defined in `suite`.

### LINKED ISSUE #645: Allow `fun` inside of `suite` blocks
It would be great for testing if we could write helper functions inside of `suite` and `test` blocks.

```mint
suite "Tests" {
  fun helperHere() {
  }

  test "it does something" {
    fun orHere() {
    }
  }
}
```

A fairly easy workaround is creating a `module` to contain your test helper functions, but that would seem largely unnecessary in most cases if we had direct `fun` support.

gdotdesign:
Fixes #645

### DIFF
--- a/core/tests/tests/Test.mint
+++ b/core/tests/tests/Test.mint
@@ -7,3 +7,23 @@ suite "Test (Async)" {
     await Test.Context.of(true)
   }
 }
+
+suite "Test (Async)" {
+  test "await Bool" {
+    await true
+  }
+
+  test "await Test.Context(Bool)" {
+    await Test.Context.of(true)
+  }
+}
+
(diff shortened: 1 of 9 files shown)


core/tests/tests/Test.mint
```diff
@@ -7,3 +7,23 @@ suite "Test (Async)" {
...
+
+suite "Test (Async)" {
+  test "await Bool" {
+    await true
+  }
+
+  test "await Test.Context(Bool)" {
+    await Test.Context.of(true)
+  }
+}
```
Sija, line 19: Duplicated tests.

```suggestion
```

Sija (approved):

PR #673, opened 2023-09-19
Test runner improvements.

### LINKED ISSUE #644: Support `mint test --watch` flag
It would be great for productivity when trying to do TDD if we could have a `mint test --watch` flag. This would automatically rerun tests when active files are modified. 

Just like [Jest](https://jestjs.io/docs/cli#--watch)

gdotdesign:
This PR adds some improvements to the `test` command and the test runner:

- The tests are compiled on the request instead of being cached, so when running in the browser and clicking refresh will load the latest tests instead of serving the cached tests. This allows for quicker iteration on tests.
- Added `--watch` for running tests when files change.

Fixes #644

### DIFF
--- a/src/test_runner.cr
+++ b/src/test_runner.cr
@@ -5,6 +5,7 @@ module Mint
     class Message
       include JSON::Serializable
 
+      property id : String
       property type : String
       property name : String?
       property suite : String?
@@ -31,8 +32,10 @@ module Mint
     @reporter : Reporter
     @browser_path : String?
     @script : String?
+    @browser : {Process, String}? = nil
 
     @failed = [] of Message
+    @test_id = Random::Secure.hex
     @succeeded = 0
 
     def initialize(@flags : Cli::Test::Flags, @arguments : Cli::Test::Arguments)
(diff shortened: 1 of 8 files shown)


src/test_runner.cr
```diff
@@ -186,13 +211,17 @@ module Mint
...
-        process.try &.signal(:kill) rescue nil
+        cleanup_browser
+      end
+    end
+
+    def cleanup_browser
+      @browser.try do |(process, profile_directory)|
+        process.signal(:kill) rescue nil
         FileUtils.rm_rf(profile_directory)
       end
```
Sija, line 226: `@browser = nil` is missing at the end of the method

src/test_runner.cr
```diff
@@ -277,26 +303,28 @@ module Mint
...
-      when "ERRORED"
-        @reporter.errored data.name, data.result
-        @failed << data
-      when "CRASHED"
-        @reporter.crashed data.result
-        @failed << data
-
-        @reporter.done
-        stop_server
+      if data.id == @test_id
```
Sija, line 306:
```suggestion
      return unless data.id == @test_id
```

src/test_runner.cr
```diff
@@ -30,9 +31,10 @@ module Mint
     @artifacts : TypeChecker::Artifacts?
     @reporter : Reporter
     @browser_path : String?
-    @script : String?
+    @browser : {Process, String}? = nil
```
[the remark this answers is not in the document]
gdotdesign: I didn't know it defaults to `nil` in this case :slightly_smiling_face:

Sija (approved):

PR #675, opened 2023-09-24
Improve Development Server

gdotdesign:
This PR improves the development server (`mint start`) in two ways:

1. Previously, it type-checked the whole application twice on every change :cold_sweat:

	- https://github.com/mint-lang/mint/blob/master/src/reactor.cr#L93
	- https://github.com/mint-lang/mint/blob/master/src/workspace.cr#L242

	Now it just uses the `Workspace` for the type-checking like everything else.

2. It just type-checks the hot path, not all entities. This speeds up type checking in cases like icons (https://github.com/mint-lang/mint-tabler-icons/blob/master/source/Icons.mint ~60K LOC) where only the icons which are used are type-checked.

After merging this, the roundtrip on every change should be twice as fast as before :tada:

### DIFF
--- a/src/workspace.cr
+++ b/src/workspace.cr
@@ -170,11 +170,11 @@ module Mint
       end
 
       spawn do
-        @env_watcher.try(&.watch do
+        @env_watcher.try &.watch do
           Env.load do
             update_cache
           end
-        end)
+        end
       end
     end
 
--- a/src/workspace.cr
+++ b/src/workspace.cr
@@ -48,6 +48,7 @@ module Mint
 
     @event_handlers = {} of String => Array(ChangeProc)
     @cache = {} of String => Ast
+    @env_watcher : Watcher?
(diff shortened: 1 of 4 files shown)


src/workspace.cr
```diff
@@ -160,6 +168,14 @@ module Mint
           reset_cache
         end
       end
+
+      spawn do
+        @env_watcher.try(&.watch do
+          Env.load do
+            update_cache
+          end
+        end)
```
Sija, line 177:
```suggestion
        @env_watcher.try &.watch do
          Env.load do
            update_cache
          end
        end
```

Sija (approved):

src/workspace.cr
gdotdesign: Oh, I thought the block would belong to the `try`, nice to know :D

PR #676, opened 2023-09-25
Extra build mode flags: `--skip-manifest`, `--inline`, and `--watch`

farism:
Adds three flags to `mint build` command:

`--skip-manifest` which will skip generating the web manifest file and reference in html.
`--inline` which will inline the JS and CSS into html, resulting in one single file for entire SPA.
`--watch` mode which will write files on changes.

### DIFF
--- a/src/builder.cr
+++ b/src/builder.cr
@@ -4,19 +4,7 @@ module Mint
       build(relative, skip_manifest, skip_service_worker, skip_icons, optimize, inline, runtime_path)
 
       if watch
-        workspace = Workspace.current
-        workspace.on "change" do |result|
-          case result
-          when Ast
-            terminal.reset
-            terminal.divider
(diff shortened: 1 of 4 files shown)


src/builder.cr
```diff
@@ -27,16 +47,18 @@ module Mint
       index_js, artifacts =
         index(json.application.css_prefix, relative, optimize, runtime_path, json.web_components)
 
-      File.write Path[DIST_DIR, "index.js"], index_js
+      if !inline
```
gdotdesign, line 50: I think it's better to group all `!inline` conditions into an outer `if`.

gdotdesign (changes requested):

Sija (changes requested):

farism:
Thanks for the review, made requested changes.

gdotdesign (approved):

Sija (approved):

PR #702, opened 2024-11-09
Allow generating source maps.

gdotdesign:
This PR builds on and supersedes #356 (Mainly the VLQ implementation is brought over).
- SPEC: https://tc39.es/source-map/
- Testing: https://evanw.github.io/source-map-visualization/ https://sokra.github.io/source-map-visualization/
- I've tested this with `mint-website` and `mint-ui-website` and it seems to be working.
- It's slow at this time (~1s from a ~2.2s build).

--------

Developer Notes:
- Because of the new compiler architecture it was relatively easy to implement.
- Changes in specs are because the order of resolving entities changed.

TODO:
- [x] Mappings and sources
- [x] Wiring to commands (`--generate-source-maps` flag)
- [x] Symbol names for the debugger 
- [x] Cleanup pass

The reason for the slowness is `Ast::Node#location` calculation from offset to line/column, which needs to be calculated when parsing so it will be in a separate PR.

### DIFF
--- a/src/utils/source_map_generator.cr
+++ b/src/utils/source_map_generator.cr
@@ -1,0 +1,110 @@
+module Mint
+  # This class is responsible for generating a source map.
+  class SourceMapGenerator
+    alias Mapping = Tuple(Ast::Node | Nil, Tuple(Int32, Int32), String | Nil)
+
+    # The line mappings.
+    getter mappings = {} of Int32 => Array(Mapping)
+
+    # The actual content of the sources.
+    getter sources_content = [] of String
+
+    # The source names for the mappings.
+    getter source_names = [] of String
+
+    # The filenames of the sources.
+    getter sources = [] of String
+
+    # The generate JavaScript file.
+    getter generated : String
+
+    def initialize(unsorted : Deque(Mapping), @generated)
+      # We sort the mappings into lines.
+      unsorted.each do |item|
+        mappings[item[1][0]] ||= [] of Mapping
+        mappings[item[1][0]] << item
+      end
+    end
+
+    def get_source_index(file)
+      sources.index(file.path) || begin
+        sources_content << file.contents
+        sources << file.path
+        sources.size - 1
+      end
+    end
+
+    def get_source_name_index(value)
+      source_names.index(value) || begin
+        source_names << value
+        source_names.size - 1
+      end
+    end
+
+    def generate : String
+      last_name_index = 0
+      last_column = 0
+      last_index = 0
+      last_line = 0
+
+      generated_mappings =
+        (0..generated.lines.size).map do |line_index|
+          if items = mappings[line_index]?
+            items.reduce({[] of String, 0}) do |(memo, last), item|
+              column =
+                item[1][1]
+
+              segment =
+                VLQ.encode(column - last)
+
+              if node = item[0]
+                # TODO: After the refactor of location remove
+                #       this temporary variable.
+                location =
+                  node.location
+
+                source_line =
+                  location.start[0] - 1
+
+                source_column =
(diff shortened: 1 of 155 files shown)


src/utils/vlq.cr
Sija, line 1: Perhaps it would make sense to extract it as a separate shard?

gdotdesign: Ideally yes, but I'm not up to do it (and keep maintaining it).

src/utils/source_map_generator.cr
```diff
@@ -0,0 +1,110 @@
...
+    getter sources = [] of String
+
+    # The generate JavaScript file.
+    getter generated : String
+
+    def initialize(unsorted : Deque(Mapping), @generated)
+      # We sort the mappings into lines.
+      unsorted.each do |item|
+        mappings[item[1][0]] ||= [] of Mapping
+        mappings[item[1][0]] << item
```
Sija, line 25:
```suggestion
        (mappings[item[1][0]] ||= [] of Mapping) << item
```

src/utils/source_map_generator.cr
```diff
@@ -0,0 +1,110 @@
...
+
+    def generate : String
+      last_name_index = 0
+      last_column = 0
+      last_index = 0
+      last_line = 0
+
+      generated_mappings =
+        (0..generated.lines.size).map do |line_index|
+          if items = mappings[line_index]?
```
Sija, line 52:
```suggestion
          next "" unless items = mappings[line_index]?
```

src/utils/source_map_generator.cr
```diff
@@ -0,0 +1,110 @@
...
+                last_index = source_index
+                last_line = source_line
+              end
+
+              memo << segment
+              {memo, column}
+            end[0].join(",")
+          else
+            ""
+          end
```
Sija, line 98:
```suggestion
```

src/utils/source_map_generator.cr
```diff
@@ -0,0 +1,110 @@
...
+                end
+
+                last_column = source_column
+                last_index = source_index
+                last_line = source_line
+              end
+
+              memo << segment
+              {memo, column}
+            end[0].join(",")
```
Sija, line 95:
```suggestion
            end.first.join(",")
```

src/utils/vlq.cr
```diff
@@ -0,0 +1,116 @@
...
+      while continue
+        char = chars.shift or raise "Expected more digits in base 64 VLQ value."
+        digit = base64_decode(char)
+        continue = false if (digit & VLQ_CONTINUATION_BIT) == 0
+        digit &= VLQ_BASE_MASK
+        vlq += digit << shift
+        shift += VLQ_BASE_SHIFT
+      end
+
+      [from_vlq_signed(vlq), String.new(chars)]
```
Sija, line 70:
```suggestion
      {from_vlq_signed(vlq), String.new(chars)}
```

src/utils/vlq.cr
```diff
@@ -0,0 +1,116 @@
...
+
+    # Decode an array of variable length quantities from the given string
+    def self.decode_array(str)
+      output = [] of Tuple(Int32, String)
+      while !str.empty?
+        int, str = decode(str)
+        output << int
+      end
+      output
+    end
```
Sija, line 81: It seems you're creating an array of tuples, yet you append ints to it?

gdotdesign: I didn't check this class, just brought it over. We don't decode anything, so this code was not even checked.

I removed the decoding part.

Sija (approved):

PR #703, opened 2024-11-12
Refactor how location of nodes are calculated.

gdotdesign:
Source map generation triggered this refactor, but it was on my radar for a while.

The location calculation was [previously done on demand](https://github.com/mint-lang/mint/blob/master/src/ast/node.cr#L99), which meant that it was a performance problem because it iterates over the lines of the source files many times. It wasn't an issue before because only the language server was using it, but now with the source maps it presented an opportunity for refactoring.

In the new version, the [parser keeps track of the location](https://github.com/mint-lang/mint/blob/node-location-refactor/src/parser.cr#L59-L72) during parsing. Positions of AST nodes now include the line and column along with the offset (that's why all node files changed).

Now, source map generation time is down and parsing time don't seem to be affected (subject project `mint-website`).

```
Without refactor
Parsing: 295.385ms
Source map generation: 891.879ms

Refactor
Parsing: 267.674ms
Source map generation: 176.489ms
```


src/ast.cr
```diff
@@ -5,9 +5,9 @@ module Mint
 
     getter unified_modules, unified_locales
 
-    def initialize(@type_definitions = [] of TypeDefinition,
-                   @operators = [] of Tuple(Int64, Int64),
-                   @keywords = [] of Tuple(Int64, Int64),
+    def initialize(@operators = [] of Tuple(Parser::Location, Parser::Location),
+                   @keywords = [] of Tuple(Parser::Location, Parser::Location),
```
Sija, line 9: Instead of a `Tuple` I'd suggest using a `record`, or at least a `NamedTuple` for readability purposes.
Otherwise IMHO it's a bit hard to understand what the values are representing.

src/ast/node.cr
```diff
@@ -1,117 +1,42 @@
...
-            column >= start[1]
-          when line == end[0] # If on the last line
-            column < end[1]
-          else
-            contains?(line)
-          end
-        end
-      end
+      property from : Parser::Location
+      property parent : Node?
```
Sija, line 5: Shouldn't they be getters instead?

src/ast/node.cr
```diff
@@ -1,117 +1,42 @@
...
-          lines << i + 1 if char == '\n'
-        end
-
-        Location.new(
-          filename: file.path,
-          start: compute_position(lines, from),
-          end: compute_position(lines, to),
-        )
+      def source
+        file.contents[from[0], to[0] - from[0]]
```
Sija, line 35: Would it make sense to memoize the return value of this method?

src/ast/node.cr
```diff
@@ -1,117 +1,42 @@
...
       end
 
-      getter location : Location do
-        Node.compute_location(file, from, to)
+      def to_tuple
```
Sija, line 38:
```suggestion
      def to_named_tuple
```

src/compilers/suite.cr
```diff
@@ -3,7 +3,13 @@ module Mint
     def compile(node : Ast::Suite) : Compiled
       compile node do
         location =
-          [Raw.new(node.location.to_json)]
+          [
+            Raw.new({
+              start:    {node.from[1], node.from[2]},
+              end:      {node.to[1], node.to[2]},
+              filename: node.file.path,
+            }.to_json),
```
Sija, line 11: That looks odd to me, why don't you just pass a values into a dedicated `Raw.new` constructor instead of serializing them into JSON?

ditto elsewhere

src/ls/folding_range.cr
```diff
@@ -47,7 +44,7 @@ module Mint
         in TypeChecker
           type_checker.artifacts.ast
             .nodes_at_path(params.text_document.path)
-            .select { |node| node.location.start[0] != node.location.end[0] }
+            .select { |node| node.from[1] != node.to[1] }
```
Sija, line 47:
```suggestion
            .reject { |node| node.from[1] == node.to[1] }
```

src/parser.cr
```diff
@@ -2,12 +2,15 @@ module Mint
   class Parser
     include Helpers
 
+    # Offset (from the start of the file), line and column.
+    alias Location = {Int64, Int64, Int64}
```
Sija, line 6: ditto re: `Tuple` vs `record`

src/ast/node.cr
gdotdesign: No `parent` is set after initialization [here](https://github.com/mint-lang/mint/blob/master/src/parser.cr#L40), and `from` is set in code actions [here](https://github.com/mint-lang/mint/blob/master/src/ls/code_actions/module_actions.cr#L17)

src/compilers/suite.cr
gdotdesign: `Raw` means that it should render the given value in JavaScript as is, it is used in many places.

src/ast/node.cr
gdotdesign: I removed it since it's not used.

src/ast/node.cr
```diff
@@ -1,117 +1,40 @@
...
-          end: compute_position(lines, to),
-        )
+      def new_line?
+        to.line > from.line
       end
 
-      getter location : Location do
-        Node.compute_location(file, from, to)
+      def source
+        @source ||= file.contents[from.offset, to.offset - from.offset]
```
Sija, line 37:
```suggestion
      getter source : String do
        file.contents[from.offset, to.offset - from.offset]
```

src/parser.cr
```diff
@@ -172,14 +184,14 @@ module Mint
...
-        input[position + i]? == char
+        input[position.offset + i]? == char
       end
     end
 
     # Consumes a word and steps the cursor forward if successful.
     def word!(expected : String) : Bool
       if word?(expected)
-        @position += expected.size
+        expected.size.times.each { step }
```
Sija, line 194:
```suggestion
        expected.size.times { step }
```

src/parser/location.cr
```diff
@@ -0,0 +1,30 @@
+module Mint
+  class Parser
+    # Offset (from the start of the file), line and column.
+    record Location, offset : Int64, line : Int64, column : Int64 do
+      def initialize
+        @offset = 0
+        @column = 0
+        @line = 1
+      end
```
Sija, line 9:
```suggestion
    record Location, offset : Int64 = 0, line : Int64 = 0, column : Int64 = 0 do
```

src/ast/node.cr
gdotdesign: TIL :+1: :smile:

PR #704, opened 2024-11-16
Improve debugging of Mint values.

### LINKED ISSUE #641: Improve debug information for complex types
It would be nice to preserve some of the debug information for complex types so that `Debug.log` would be more useful. This would only take effect when developing, not when doing a production (minified) bundle.

Some examples are below (generated names can be differnet)

```mint
record Point2D {
  x : Number,
  y : Number
}

component Main {
  fun foo {
    ""
  }

  fun componentDidMount {
    /*
    Prints `r {x:0, y:0}`

    It would be nice if this was `Point2D {x:0, y:0}`
    */
    Debug.log(Point2D(0, 0))

    /*
    Prints an object `AI {_0: 'hello world', length: 1}`

    It would be nice if it would print the as-is `Maybe::Just("hello world")`
    */
    Debug.log(Maybe::Just("hello world"))

    /*
    Prints `ƒ a() { return ''; }`

    It would be nice if this set the function name so it would print `ƒ foo() { return ''; }`
    */
    Debug.log(foo)
  }

  fun render : Html {
    <div/>
  }
}
```

gdotdesign:
This PR improves how Mint values (types) are printed. Before this PR, when using `Debug.log` or `dbg` the JavaScript versions of values are being printed to the console:

image

Which is OK, but we can do better. This PR changes that to this:

image

Basically, we instrument the generated code, injecting the name of the records and variants so we can pretty print them.

If we are using `dbg` we have a location as well (this is not new):

image

Other things:

- Added `Debug.inspect` which returns the pretty printed version.
- `Debug.log` remains as-is printing the raw values.
- `dbg` is now using the pretty printed version by default, to get the raw value we can use `dbg!`
- The location line is not printed if source maps are enabled because it's redundant since it will point to the source.
  image

------

Resolves #641.


runtime/src/debug.js
```diff
@@ -0,0 +1,128 @@
+import indentString from "indent-string";
+import { isVnode } from './equality';
+import { Variant } from './variant';
+import { Name } from './symbols';
```
Sija, line 4: Use either single or double quotes consistently.
Even better would be to configure ESLint/Prettier to do it automatically.

ditto elsewhere

src/compilers/dbg.cr
```diff
@@ -8,20 +8,34 @@ module Mint
...
+          end
+
         if expression = node.expression
           js.iif do
             js.statements([
               js.const(var, compile(expression)),
-              js.call(["console.log"] of Item, [location]),
-              js.call(["console.log"] of Item, [var]),
+              location,
+              js.call(["console.log"] of Item, [arg]),
```
Sija, line 30: Why not log `location` in the same `console.log` call as the inspected expression?

```suggestion
              js.call(["console.log"] of Item, [location, arg]),
```

src/compilers/dbg.cr
```diff
@@ -8,20 +8,34 @@ module Mint
...
               js.return(var),
             ])
           end
         else
           js.arrow_function([var]) do
             js.statements([
-              js.call(["console.log"] of Item, [location]),
-              js.call(["console.log"] of Item, [var]),
+              location,
+              js.call(["console.log"] of Item, [arg]),
```
Sija, line 38: ditto

```suggestion
              js.call(["console.log"] of Item, [location, arg]),
```

src/compilers/dbg.cr
gdotdesign: They will be printed on the same line:

image

runtime/src/debug.js
gdotdesign: We have Prettier set up https://github.com/mint-lang/mint/blob/master/runtime/Makefile#L21 but at this time it would produce many unrelated changes, so it would be better to run it idependetly in an other commit.

Sija (approved):

PR #705, opened 2024-11-16
Added position related APIs.

### LINKED ISSUE #498: Include/Implement Dom.offsetLeft or Html.Event.layerX or Html.Event.offsetX
i) offsetLeft / offsetRight
ii) layerX / layerY
iii) offsetX / offsetY 
These three things really helps in DOM manipulation. It is very cumbersome to do things without these methods. It will bring developer happiness hopefully.

gdotdesign:
Added:

- `Dom.offsetLeft`, `Dom.offsetTop` methods
- `Html.Event.layerX`, `Html.Event.layerY` fields
- `Html.Event.offsetX`, `Html.Event.offsetY` fields

Fixes #498


core/source/Dom.mint
```diff
@@ -590,4 +590,28 @@ module Dom {
...
+  /*
+  Returns the distance from the outer border of the element (including its
+  margin) to the left padding edge of the closest positioned ancestor element.
+
+    if let Just(div) = Document.getElementBySelector("div") {
+      Dom.offsetLeft(div)
+    }
+  */
+  fun offsetLeft (element : Dom.Element) : Number {
+    `#{element}.offsetLeft || -1`
```
Sija, line 603: Returning `-1` is here is IMO totally unexpected and may lead to hard-to-debug bugs.
Also, AFAIU it's not coherent with this method documentation.

ditto below

gdotdesign: We have other functions that return values similarly https://github.com/mint-lang/mint/blob/master/core/source/Dom.mint#L371 https://github.com/mint-lang/mint/blob/dom-position-things/runtime/src/normalize_event.js#L93 We could return `Maybe(Number)` (and for all similar functions) but it would greatly increase the friction when using this.

I'll update the documentation comment to reflect this.

gdotdesign: If the element is not in the DOM, it returns `0`:
```
document.createElement("div").offsetLeft
```

So maybe returning `0` if it's not applicable is better?

Sija: `0` sounds better to me, although it still seems more like a hack tbh...

Sija: I guess it's good for now, yet it needs to be changed for other props as well.

runtime/src/normalize_event.js
```diff
@@ -102,6 +102,14 @@ export const normalizeEvent = (event) => {
...
           case "screenY":
             return -1;
+          case "layerX":
+            return -1;
+          case "layerY":
+            return -1;
+          case "offsetX":
+            return -1;
+          case "offsetY":
+            return -1;
```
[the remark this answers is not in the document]
gdotdesign: https://github.com/mint-lang/mint/commit/ed38e9f122c7a275b44dd09d6c3fd196ee8c04fb

Issue #397, opened 2021-04-15
Error parsing `v+1` expression (with no spaces around +)


cdevienne:
Given the following enum: 
```
enum Counter {
    Value(Number)
    NoValue
}
```

I wrote the following code in a component:
```
  state counter : Counter = Counter::NoValue

  fun inc {
      case (counter) {
        Counter::Value v =>
          next { counter = Counter::Value(v+1) }
        Counter::NoValue =>
          Promise.never()
      }
  }
```

The message I get from the compiler is:
```
The body of a record must end with a closing bracket.

I was looking for the bracket } but found ounter::Value(v+1) instead. 
```

It took me a while to realize that the error came from the absence of whitespaces around the "+" token.

I think the ideal solution would be to accept "v+1" as a valid expression equivalent to "v + 1", but at least we should have a less misleading error message.

PR #706, opened 2024-11-17
Relax whitespace requirement around operators.

### LINKED ISSUE #637: Relax whitespace rules around operators
The compiler could allow all of the following syntax

```
let a=2
let b = a|>Number.toString
let neq = a!=1
let eq = foo==2
let lte = a<=3
let lt = a<3
let gte = a>=1
let gt = a>1
let plus = a+1
let multiply = a*2
let divide = a/2
let mod = a%2==0
let exp = 2**3
let and = a==2&&a==2
let ortest = a==1||a==2
let true = ! false
```

Special consideration needs to be taken for 

`-` because `-` is ambiguous with negative values
`or` because it the characters "or" can be used in variable names

In addition to relaxing the syntax, if there is no whitespace surrounding an operator, then whitespace should be inserted by the formatter. Not sure if that should be a separate issue.

gdotdesign:
Fixes #637 #397

### DIFF
--- a/src/parser/top_level.cr
+++ b/src/parser/top_level.cr
@@ -79,7 +79,6 @@ module Mint
             component ||
             constant ||
             property ||
-            operator ||
             provider ||
             function ||
             comment ||
--- a/spec/examples/pipe
+++ b/spec/examples/pipe
@@ -64,3 +64,15 @@ component Main {
     ""
   }
 }
+-------------------------------------------------------------------------------
+component Main {
+  fun test (value : String) {
+    value
+  }
+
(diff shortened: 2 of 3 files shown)


src/parser/top_level.cr
```diff
@@ -79,7 +79,6 @@ module Mint
             component ||
             constant ||
             property ||
-            operator ||
```
gdotdesign, line 82: Operators are parsed in the last part in `expression`.

Sija (approved):

PR #708, opened 2024-11-21
Use relative path for error messages.

gdotdesign:
This PR changes the paths of the snippets in error messages to be relative to the project root - the directory where `mint.json` file is. This makes it easier to find since we don't need to read the whole path.

From this:
```
   ┌ /path/to/project/source/Main.mint:3:7
   ├──────────────────────────────────────
```

To this:
```
   ┌ source/Main.mint:3:7
   ├──────────────────────────────────────
```

--------------------------

The only relevant change is in `src/errorable.cr`, the other ones are just snapshot updates for tests.

### DIFF
--- a/spec/errors/access_expected_field
+++ b/spec/errors/access_expected_field
@@ -6,8 +6,8 @@ component Main {
 
 I was expecting the name of the accessed entity but I found "a space" instead:
 
-   ┌ ./spec/errors/access_expected_field:3:7
-   ├────────────────────────────────────────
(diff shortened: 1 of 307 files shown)


Sija (commented):
I can't find the changed code in question because of the massive diff from the snapshots. Could you please split the commit in two - one with the change itself and one with the snapshot updates?

gdotdesign:
Pretty small diff: `value.file.path` -> `value.file.relative_path`

image

Sija (approved):

Issue #709, opened 2024-11-21
Unable to install mint-ui with current instructions and mint 0.20.0


jhf:
I'm reporting it here since I think it is related to the `mint install` command handling of ssh vs https.

After `mint init` I add 
```
"dependencies": {
  "mint-ui": {
    "repository": "https://github.com/mint-lang/mint-ui",
    "constraint": "8.0.0 <= v < 9.0.0"
  }
}
```
as per the instructions, and then I run `mint install` and the installation fails with
```
mint install
Mint - Installing dependencies
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
⚙ Constructing dependency tree...

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

░ ERROR (REPOSITORY_COULD_NOT_UPDATE) ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░

I could not update the repository: https://github.com/mint-lang/mint-ui.git

The error I got from the git command is this:

fatal: not a git repository (or any of the parent directories): .git

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
There was an error, exiting...
```

However if I change my dependencies to
```
  "dependencies": {
    "mint-ui": {
      "repository": "git@github.com:mint-lang/mint-ui.git",
      "constraint": "8.0.0 <= v < 9.0.0"
    }
  }
```
then `mint install` is able to resolve mint-ui, but not mint-color, as you see here.
```
mint install
Mint - Installing dependencies
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
⚙ Constructing dependency tree...
  ✔ Updated mint-ui (git@github.com:mint-lang/mint-ui.git)

⚙ Resolving dependency tree...

[...]

PR #712, opened 2024-11-21
Feature/git error hint

jhf:
Provide a hint for the user to run a cleanup to avoid the errors in https://github.com/mint-lang/mint/issues/709 if the package state is corrupt for some reason.

### DIFF
--- a/src/installer/repository.cr
+++ b/src/installer/repository.cr
@@ -63,6 +63,12 @@ module Mint
           block do
             bold error.to_s.strip
           end
+
+          block do
+            text "Hint: Run "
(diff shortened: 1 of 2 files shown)


gdotdesign (approved):

src/installer/repository.cr
```diff
@@ -63,6 +63,12 @@ module Mint
           block do
             bold error.to_s.strip
           end
+
+          block do
+            text "Hint: Run "
+            bold "mint tool clean; mint tool clean --package-cache"
+            text " to reset local state, and then try again."
```
Sija, line 70: Extraneous space here.

ditto elsewhere

PR #721, opened 2024-11-27
Highlight single character errors.

### LINKED ISSUE #717: Missing space before colon after function parameter
Hi!

By the example here:

```cpp
  fun mapToObjectError (result : Result(String, a)) : Result(Object.Error, a) {
    Result.mapError(result, (err: String) { Object.Error(err) })
  }
```

The `err: String` is missing a space before colon and doesn't show where:

Image

gdotdesign:
Fixes #717 Also contains a change related to this: connect variable eats trailing whitespace if there is no `as name` part, which with the changes in the PR would highlight the whitespace.

### DIFF
--- a/src/utils/terminal_snippet.cr
+++ b/src/utils/terminal_snippet.cr
@@ -34,16 +34,16 @@ module Mint
           self[0, diff_from]
 
         center =
-          self[diff_from, diff_to].colorize.on(:white).fore(:red).to_s
+          self[diff_from, diff_to]
 
         right =
           self[diff_to, contents.size]
 
         highlighted =
-          left + center + right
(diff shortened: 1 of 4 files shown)


src/utils/terminal_snippet.cr
```diff
@@ -34,16 +34,16 @@ module Mint
...
 
         right =
           self[diff_to, contents.size]
 
         highlighted =
-          left + center + right
+          left + center.colorize.on(:white).fore(:red).to_s + right
 
         arrows =
-          (" " * left.size) + ("⌃" * center.uncolorize.size)
```
gdotdesign, line 46: `uncolorize` removes the whitespace, which resulted in cases where nothing was highlighted.

Sija (approved):

PR #738, opened 2025-02-20
Fix typo when unable to find file

EvanHahn:
`Ast::Directives::FileBased` had a typo in the error message when it couldn't find a file. This fixes that.


gdotdesign:
Thank you :pray:

PR #749, opened 2025-03-17
Implement function captures.

gdotdesign:
This PR implements function captures, which is a way to have partial application:

```mint
fun add(a: Number, b: Number) : Number {
  a + b
}

// All of them are the same:
let addOne = (c: Number) { add(c, 1) }
let addOne = add(a: _, b: 1)
let addOne = add(_, 1)
```


Sija (approved):

ryanprior:
Can't wait 😤

PR #750, opened 2025-03-23
Automatically serialize custom types

gdotdesign:
This PR adds the ability to automatically serialize custom types into an opinionated format.

Most of the time, developers don't really care what format the data is in, they just want to persist it somewhere (like local storage) and load it back. Right now, they have to lean about custom decoders and encoders if the data is not a record (we have automatic serialization for those) and that's a roadblock for that use case.

After this, `encode` and `decode` will work on variant types:

```mint
type User {
  LoggedIn(name : String, age: Number)
  LoggedOut
}

let loggedInUser =
  User.LoggedIn(name: "Joe", age: 38)

// Encodes it to an object: { type: "User.LoggedIn", value: ["Joe", 38] }
let encoded =
  encode loggedInUser 

(decode encoded as User) == loggedInUser
```

The serialized structure is:
```
{
  type: ...
  value: [...]
}
```
- `type` is the fully qualified name of the variant
- `value` contain the parameters in defined order

--- 

The PR also contains:
- A fix for `styles` helper function
- A fix for normalize event tests
- Formatting of the runtime files
- Test for the runtime decoders


Sija:
In vast majority of the cases, you don't control the backend, so the serialization format must be universal and even if you do, you don't want to tailor it for the specific frontend and its serialization format. Therefore I'd suggest to rewrite this into a general `key: value` format, which should cover majority of the use cases.

Issue #752, opened 2025-03-25
`Array.all` is missing


hamidb80:
Hey, 

since some other functional programming language implemented `all` along with `any` function ([like in Elixir](https://hexdocs.pm/elixir/1.18.3/Enum.html#all?/2)], it is expected to see `all` function in `Array` module in the documentation of Mint programming language.

PR #753, opened 2025-03-26
Adds `Array.all` function

ryanprior:
Fixes https://github.com/mint-lang/mint/issues/752

- new function: Array.all
- new test suite for the same

### DIFF
--- a/core/source/Array.mint
+++ b/core/source/Array.mint
@@ -16,6 +16,17 @@ module Array {
   }
 
   /*
+  Returns `true` if all items in the array matche the predicate function,
+    `false` otherwise.
+
+    Array.all([1, 2, 3, 4], (number : Number) { number % 2 == 0 }) == false
+    Array.all(["hello", "mint"], (str : String) { String.size(str) > 3 }) == true
+  */
+  fun all (array : Array(item), function : Function(item, Bool)) : Bool {
+    reduce(array, true, (memo : Bool, val : item) { memo && function(val) })
+  }
+
+  /*
   Merges two arrays together into a new one.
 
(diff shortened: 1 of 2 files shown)


core/source/Array.mint
```diff
@@ -15,6 +15,17 @@ module Array {
...
 
+  /*
+  Returns `true` if all items in the array matche the predicate function,
+    `false` otherwise.
+
+    Array.all([1, 2, 3, 4], (number : Number) { number % 2 == 0 }) == false
+    Array.all(["hello", "mint"], (str : String) { String.size(str) > 3 }) == true
+  */
+  fun all (array : Array(item), function : Function(item, Bool)) : Bool {
+    reduce(array, true, (memo : Bool, val : item) { memo && function(val) })
```
ryanprior, line 26: Design decision: does the empty array always satisfy "all" or never satisfy "all"? In Crystal the answer is that it always satisfies "all" (ref: https://play.crystal-lang.org/#/r/htfj) so I made this behavior match.

gdotdesign: It should return true for empty arrays: https://en.wikipedia.org/wiki/Vacuous_truth#In_computer_programming :+1:

core/source/Array.mint
```diff
@@ -15,6 +15,17 @@ module Array {
     Array.find(array, function) != Maybe.Nothing
   }
 
+  /*
+  Returns `true` if all items in the array matche the predicate function,
```
gdotdesign, line 19:
```suggestion
  Returns `true` if all items in the array match the predicate function,
```

gdotdesign (approved):

gdotdesign:
Thank you  :pray:

PR #757, opened 2025-03-28
Alternative Patterns

gdotdesign:
This PR adds alternative patterns to case statements.

```mint
case number {
  0 | 1 | 2 | 3 => "0-3"
  _ => "other"
}
```


src/type_checkers/case.cr
```diff
@@ -106,6 +106,24 @@ module Mint
...
+          ExhaustivenessChecker::POr.new(
+            case memo
+            when ExhaustivenessChecker::POr
+              memo
+            else
+              to_pattern(memo)
+            end,
+            to_pattern(node)
+          )
+        end].select(ExhaustivenessChecker::Pattern).first
```
Sija, line 123:
```suggestion
        end].first(&.is_a?(ExhaustivenessChecker::Pattern))
```

PR #770, opened 2025-05-10
Add command line switch to display browser output during mint test

MathiusD:
In some case (e.g. during CI build for example), it can be useful to show output of browser invoked by mint (For example when browser crash or exits with an error due of some configuration error)

So i propose to add following command line switch (--show-browser-output) for this purpose


gdotdesign (approved):

gdotdesign:
Thank you :pray:

PR #822, opened 2026-05-09
Migrate from yarn to bun

Sija:


ryanprior:
Have you been following the recent drastic changes to the bun codebase? Might see how that shakes out before pulling the trigger on a migration.

gdotdesign:
Yes, I saw. Not sure why would we want to change the runtime @Sija?

Sija:
I haven't been aware of those tbh, just catching up and it... doesn't look good 😬

